PR #158, opened 2016-11-16
C matrices

JoostJM:
Implement C for calculation of matrices
Currently supports:
- GLCM
- GLSZM

This greatly improves processing speed, while still keeping an all-python option available  (currently enabled by setting `radiomics.debugging` to `True`.

Also included automatic linking and compiling to setup script. Needs additional package (`cython`) to work, but is now also included in setup script (if not installed, setup script installs it without error).

Lastly, added a new test, where the python generated matrices are compared to the C generated matrices for all test cases. No features are calculated here, but matrices are assessed element wise (maximum difference for any element is 1e-3, this is due to small machine precision errors possible in NGTDM).


JoostJM:
@jcfr I updated my scikit-build and cmake on my computer at work, but I am still getting errors. It's an unresolved external reference error in the load-function of numpy (only fails when compiling with cmake though). Here is a section the setup output, do you know whats going wrong?

image

jcfr:
@JoostJM  Note that I started to rebase and re-organized commits. I will follow up with some update.

jcfr:
@JoostJM  Here is what I have so far: https://github.com/Radiomics/pyradiomics/tree/c-matrices-rebased-jcfr

I squashed together changes that made sense. Still need some work.

Would be nice to integrate separately all changes that are not specific to C implementation. 

For example: https://github.com/Radiomics/pyradiomics/commit/f6b77ed991095240a44af0e9d40fbe2e166e7ffe

jcfr:
Closing. This is superseded by #200

PR #175, opened 2017-01-04
fix incorrectly hardcoded label when calculating volume

haarburger:
This fixes an incorrectly hardcoded label that causes the calculation of features for labels != 1 to fail.

@JoostJM, unfortunately, I can't mention the whole team as suggested in the contribution guidelines.


JoostJM:
@haarburger Thanks for the contribution!
As to mentioning the whole team, did the @Radiomics/developers not work?

haarburger:
Your'e welcome! The framework is awesome :+1: 

I can mention contributors individually, but if I try to mention the team, no automatic suggestion shows up. Maybe I did something wrong? Or is this only possible for team members? I'm not that familiar with github.

PR #182, opened 2017-01-17
Fix incorrect slicing during application of wavelet filter

JoostJM:
Due to some slicing errors, the wavelet filter resulted in a 90deg rotation and a mirroring along one axis. This resulted in that the labelmaps did not align with the ROI anymore, causing the wavelet features to be incorrectly calculated.

This PR addresses this error by recoding the slicing during wavelet decomposition. Results have been checked visually. I'll add some examples later.

### DIFF
--- a/radiomics/imageoperations.py
+++ b/radiomics/imageoperations.py
@@ -251,13 +251,13 @@ def applyThreshold(inputImage, lowerThreshold, upperThreshold, insideValue=None,
 def swt3(inputImage, wavelet="coif1", level=1, start_level=0):
   matrix = sitk.GetArrayFromImage(inputImage)
   matrix = numpy.asarray(matrix)
-  data = matrix.copy()
-  if data.ndim != 3:
+  if matrix.ndim != 3:
     raise ValueError("Expected 3D data array")
 
   original_shape = matrix.shape
   adjusted_shape = tuple([dim + 1 if dim % 2 != 0 else dim for dim in original_shape])
-  data = numpy.resize(data, adjusted_shape)
+  data = matrix.copy()
+  data.resize(adjusted_shape, refcheck=False)
 
   if not isinstance(wavelet, pywt.Wavelet):
     wavelet = pywt.Wavelet(wavelet)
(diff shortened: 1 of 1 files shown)


JoostJM:
Original

original

LLL: Image is flipped and rotated --> This is due to incorrect slicing

wavelet_lll_old

LLH: Bottom slice is somehow 'copied' / filtered and inserted over the top slice --> This is due to the wavelet transform by the pywavelet package, will look into this further...

wavelet_llh_old

JoostJM:
Tests of the pywavelet functions (using Haar wavelet): 

Test1, length = 8
In: [5 4 3 0 0 0 0 0]
Out-H: [ 6.364   4.9497  2.1213  0.      0.      0.      0.      3.5355]
Out-L: [ 0.7071  0.7071  2.1213  0.      0.      0.      0.     -3.5355]

Test2, length = 6
In: [5 4 3 0 0 0]
Out-H: [ 6.364   4.9497  2.1213  0.      0.      3.5355]
Out-L: [ 0.7071  0.7071  2.1213  0.      0.     -3.5355]

Test3, reverse
In: [0 0 0 0 0 3 4 5]
Out-H: [ 0.      0.      0.      0.      2.1213  4.9497  6.364   3.5355]
Out-L: [ 0.      0.      0.      0.     -2.1213 -0.7071 -0.7071  3.5355]

Test4, reverse with extra padding
In: [0 0 0 0 0 3 4 5 0 0]
Out-H: [ 0.      0.      0.      0.      2.1213  4.9497  6.364   3.5355  0.      0.    ]
Out-L: [ 0.      0.      0.      0.     -2.1213 -0.7071 -0.7071  3.5355  0.      0.    ]

Test5, discreet instead of stationary wavelet transform
In: [5 4 3 0 0 0 0 0]
Out-H: [ 6.364   2.1213  0.      0.    ]
Out-L: [ 0.7071  2.1213  0.      0.    ]

The value of 3.5355 is equal to 1/sqrt(2) * (5 + 0), i.e. looks like it wraps around for the last index:
1/sqrt(2) * (index[-1] + index[0]) for L, 1/sqrt(2) * (index[-1] + index[0]) for H

JoostJM:
Got a response from pywavelets: This duplication 'error' is intentional, as it is their way of handling edges: swt is forced to `periodization`, which appends copies of the array.
Furthermore, in swt, the wavelets overlap (i.e. first index in Haar is calculated using inidces 0 and 1, 2nd using 1 and 2, etc.), but with dwt there is no overlap. If I understand correctly, this is different from just upsampling the filter coefficients, correct?

I will look at the matlab code to see if I can find how the wavelets were calculated before...

PR #184, opened 2017-01-22
Add CLI scripts to setup

JoostJM:
Rename the scripts to pyradiomics and pyradiomicsbatch (easy to remeber, but unique to package, for use in the commandline). Additionally, move them to separate folder in the package, and add them to the setup script so the CLI gets distributed with the installation, not just the git repo.
Additionally, add the use of these scripts to the `Usage` section of the documentation and adapt this section further to represent the more standard use of the toolbox, including interactive extraction using the `featureextractor` module.

Finally, adapt the examples in the bin folder to make use of the example data stored in the repository, this removes the need to install the example data when pyradiomics is installed.

cc @Radiomics/developers

### DIFF
--- a/bin/helloFeatureClass.py
+++ b/bin/helloFeatureClass.py
@@ -6,11 +6,10 @@ from radiomics import firstorder, glcm, imageoperations, shape, glrlm, glszm
 # testBinWidth = 25 this is the default bin size
 # testResampledPixelSpacing = [3,3,3] no resampling for now.
 
-dataDir = os.path.abspath("") + os.path.sep + ".." + os.path.sep + "data"  # assumes this file in in pyradiomics/bin
-# imageName = str(dataDir + os.path.sep + 'prostate_phantom_subvolume.nrrd')
-# maskName = str(dataDir + os.path.sep + 'prostate_phantom_subvolume-label.nrrd')
-imageName = str(dataDir + os.path.sep + 'breast1_image.nrrd')
-maskName = str(dataDir + os.path.sep + 'breast1_label.nrrd')
+testCase = 'brain1'
+dataDir = os.path.join(os.path.abspath(""), "..", "data")
+imageName = os.path.join(dataDir, testCase + '_image.nrrd')
+maskName = os.path.join(dataDir, testCase + '_label.nrrd')
 
 if not os.path.exists(imageName):
   print 'Error: problem finding input image', imageName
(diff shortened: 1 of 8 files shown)


docs/usage.rst
```diff
@@ -6,15 +6,95 @@ Usage
...
+
+* Both scripts can be run directly from a command line window, anywhere in your system.
+
+* To extract features from a single image and segmentation run::
+
+    pyradiomics <path/to/image> <path/to/segmentation>
+
+* To extract features from a batch run::
+
+    pyradiomicsbatch <path/to/input> <path/to/output>
```
fedorov: did you mean to say `python commandline.py` and `python commandlinebatch.py`?

fedorov: No, I see it works. Nice!

bin/helloFeatureClass.py
```diff
@@ -4,9 +4,9 @@
 from radiomics import firstorder, glcm, imageoperations, shape, glrlm, glszm
 
 # testBinWidth = 25 this is the default bin size
-# testResampledPixelSpacing = (3,3,3) no resampling for now.
+# testResampledPixelSpacing = [3,3,3] no resampling for now.
 
-dataDir = os.path.dirname(os.path.abspath(__file__)) + os.path.sep + ".." + os.path.sep + "data"
+dataDir = os.path.abspath("") + os.path.sep + ".." + os.path.sep + "data"  # assumes this file in in pyradiomics/bin
```
fedorov: fyi, you could use `os.path.join` to make this prettier!

fedorov:
Maybe you can also change in this PR to use `w` mode for `argparse.FileType` in commandline.py?

https://github.com/Radiomics/pyradiomics/blob/master/bin/commandline.py#L16

bin/helloFeatureClass.py
JoostJM: @fedorov, good idea, I've been using that here and there, will change before merging!

docs/usage.rst
JoostJM: Trick is how it's defined in the setupscript. There you link your CLI command (e.g. pyradiomics) to the function it should run (in this case, `radiomics.scripts.commandline.main()`), similar for pyradiomics batch. You could also just specify the scripts as such, however, this doesn't allow you to name them differently and it requires the user to add .py in the command window. I thought it more intuitive to hava a command like `pyradiomics`

fedorov: It's a great idea, I just didn't know about this capability. I like it!

JoostJM:
Additionally, I'd like to mark it v1.0.1 after this PR and #185 are merged, as it contains added functionality (CLI now installed with the toolbox) and has some significant changes in the documentation.

On that note, do we want a separate file for the changelog added to the repository or use github's release notes? The last option feels also like a changelog...

fedorov:
I don't see a need for something in addition to the github release notes

PR #185, opened 2017-01-23
Update documentation

JoostJM:
Add some additional references in GLRLM.
Fix some errors in documentation (old names, spelling errors).
Insert active links where documentation refers to other parts of the toolbox.
Add installation instructions for Windows.
Add documentation on some functions in imageoperations.

Update requirements.txt to use PyKwalify>=1.6.0, as this contains a fix for the bug which was breaking the output in jupyter. Remove hack around this bug in the helloRadiomics notebook example.

Move application of filters to `imageoperations` (instead of having two parts, one in `featureextractor`, one in `imageoperations`). Returns filtered image as part of a python generator. Apply changes to parts of pyradiomics that uses this functionality (mainly in `featureextractor.execute` and the helloFeatureClass examples).
N.B. This change causes the potential filters to have capitals in the name, similar to enabling individual features in a feature class. Changed the parameter schema accordingly.

cc @Radiomics/developers


README.md
```diff
@@ -102,3 +96,14 @@ This package is covered by the [3D Slicer License](LICENSE.txt).
...
 <sup>6</sup>Isomics
+
+### General references
+
+- HJWL Aerts, ER Velazquez, RTH Leijenaar, et al., "Decoding tumour phenotype by noninvasive imaging using a 
+  quantitative radiomics approach", vol. 5, Nat Communication, 2014. 
+  Available [here](http://www.nature.com/ncomms/2014/140603/ncomms5006/full/ncomms5006.html).
+  Specifically, the formulation of the individual feature calculation is covered in this 
+  [supplement](http://www.nature.com/ncomms/2014/140603/ncomms5006/extref/ncomms5006-s1.pdf).
+- Zwanenburg A, Leger S, Vallières M, Löck S., "Image biomarker standardisation initiative - feature definitions",
```
fedorov: Nice reference! How about we define correspondence between the pyradiomics features and the ones defined by this document?

JoostJM: Let's discuss it this Wednesday at the call.

JoostJM: @fedorov I made this reference slightly different in the first commit in this PR, let me know what you think of that version.

JoostJM: @hugo?

fedorov: @hugoaerts

radiomics/firstorder.py
```diff
@@ -283,10 +284,11 @@ def getUniformityFeatureValue(self):
     where a greater uniformity implies a greater heterogeneity or a
     greater range of discrete intensity values.
     """
+    eps = numpy.spacing(1)
 
     bins = imageoperations.getHistogram(self.binWidth, self.targetVoxelArray)[0]
     try:
-      bins = bins / (float(bins.sum()))
+      bins = bins / (float(bins.sum() + eps))
```
fedorov: I don't see how this is better than setting the result to NaN.

radiomics/imageoperations.py
```diff
@@ -248,7 +239,117 @@ def applyThreshold(inputImage, lowerThreshold, upperThreshold, insideValue=None,
   return tif.Execute(inputImage)
 
 
-def swt3(inputImage, wavelet="coif1", level=1, start_level=0):
+def applyFilterOriginal(inputImage, **kwargs):
```
fedorov: As a historical perspective, @hugoaerts specifically didn't like the word "filter" as applied to this kind of operations, and that is why the name of the module is `imageoperations`. I personally agree that "filter" is a more recognizable and conventionally used term.

JoostJM: I can change the signature of the name quite easily, but it'll have to be a signature in common for all filters and the original image. How about "get<filtername>Image"? This doesn't specify filter and as a bonus is quite like how we define the feature function signature...

radiomics/firstorder.py
JoostJM: Both approaches have merit in my opion. I propose to leave this like it is for now and address the issue in a next hangout call. I'd like to handle these situations in one way throughout the toolbox.

fedorov: I completely agree it should be handled uniformly, and also there should be some place in the documentation where the behavior is described. I personally prefer explicit NaN. I agree let's discuss at the next call.

radiomics/imageoperations.py
fedorov: Either way is fine with me, I just wanted to make sure Hugo is happy with the conventions, and they are consistent throughout the toolkit and documentation.

PR #190, opened 2017-02-10
Move getFeatureClasses and getInputImages to radiomics namespace

JoostJM:
Move the `getFeatureClasses` and `getInputImages` out of the RadiomicsFeaturesExtractor class and to the radiomics namespace.
This makes these functions more globally available (as it is also used in the testing, which does not necessarily use the featureextractor).

Additionally, add new parameter, "additionalInfo" (Boolean) to control wheter or not to append the extra information on the extraction to the output. This additional information is provided by the generalinfo.py module.

### DIFF
--- a/radiomics/__init__.py
+++ b/radiomics/__init__.py
@@ -4,7 +4,12 @@ if sys.version_info < (2, 6, 0):
   raise ImportError("pyradiomics > 0.9.7 requires python 2.6 or later")
 in_py3 = sys.version_info[0] > 2
 
+import pkgutil
+import inspect
+import os
 import logging
+
+from . import base, imageoperations
 
 
 def debug(debug_on=True):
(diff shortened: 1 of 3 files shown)


radiomics/__init__.py
```diff
@@ -32,6 +37,54 @@ def debug(debug_on=True):
...
+      attributes = inspect.getmembers(module, inspect.isclass)
+      for a in attributes:
+        if a[0].startswith('Radiomics'):
+          if base.RadiomicsFeaturesBase in inspect.getmro(a[1])[1:]:
+            _featureClasses[mod] = a[1]
+
+  return _featureClasses
+
+
+def getInputImages():
```
fedorov: To me this name is quite confusing. Wouldn't `getInputImageTypes` make more sense?

radiomics/featureextractor.py
```diff
@@ -420,38 +422,6 @@ def computeFeatures(self, image, mask, inputImageName, **kwargs):
 
     return featureVector
 
-  @classmethod
-  def getInputImageTypes(cls):
```
fedorov: Interesting - so it *was* `getInputImageTypes` before! :)

JoostJM: ;-) Let me revert it back then.

PR #194, opened 2017-02-14
making it python 3 compatible_REVISED

JoostJM:
> without breaking the python 2 compatibility (#178), adding the future imports to keep others clean and behavior particularly division consistent.
> 
> The current changes should change or break nothing with python 2 execution and only make the code itself runnable on 3.

supersedes #188 

Additional changes to previous PR:

- Rebased on latest master reflecting changes by #190 and #191
- Re-enabled py3k support in continuous testing

cc @Radiomics/developers, @kmader

PR #195, opened 2017-02-14
STYL: testUtils: Improve test outputs

jcfr:
* Update feature name in testing output to reflect the feature class.

* write "N/A" string when no data can be found

* write csv as text file using `w` instead of `wb` (this should help with
  the transition to py3)

### DIFF
--- a/tests/testUtils.py
+++ b/tests/testUtils.py
@@ -191,20 +191,21 @@ class RadiomicsTestUtils:
     Use utility methods to get and test the results against the expected baseline value for this key.
     """
 
+    longName = '%s_%s' % (self._featureClassName, featureName)
     if value is None:
-      self._diffs[self._testCase][featureName] = None
-      self._results[self._testCase][featureName] = None
+      self._diffs[self._testCase][longName] = None
+      self._results[self._testCase][longName] = None
     assert (value is not None)
 
     if math.isnan(value):
-      self._diffs[self._testCase][featureName] = numpy.nan
-      self._results[self._testCase][featureName] = numpy.nan
+      self._diffs[self._testCase][longName] = numpy.nan
+      self._results[self._testCase][longName] = numpy.nan
     assert (not math.isnan(value))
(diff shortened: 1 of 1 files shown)


tests/testUtils.py
```diff
@@ -244,18 +245,18 @@ def writeCSV(self, data, fileName):
 
     {'id1' : {'f1':n1, 'f2':n2}, 'id2' : {'f1':n3, 'f2':n4}}
     """
-    csvFile = open(fileName, 'wb')
+    csvFile = open(fileName, 'w')
     csvFileWriter = csv.writer(csvFile)
     # get the headers from the first row
-    header = sorted(data[data.keys()[0]].keys())
+    header = list(data[list(data.keys())[0]].keys())
```
jcfr: @JoostJM This change was grabbed from 4f380d6 in #194 , is this expected ? I guess it simply create the CSV column in the order of the keys.

JoostJM: @jcfr I changed this as I forgot to add the feature class to the output (meaning the features from different classes ended up mixed together due to the sorting). However, the output now again contains also the feature classes, meaning applying the sorted() function nicely sorts the features alphabetically per featureclass (as feature classes are also sorted alphabetically). So in this case, I'd propose to revert this to `sorted(data[list(data.keys())[0]].keys())`

JoostJM: @jcfr, I'm rewriting this a bit, as it is a bit buggy if you e.g. only test the 'breast1' test case. In that case, no results are written as the first test case ('brain1') does not contain any calculated features. The test cases are based on the baseline. I'll rewrite it so, that only the test cases that have been actually tested are written to the result and diff files, with no files written if no tests were actually run.

jcfr: Makes sense. Glad I asked :smile: 

Here is a PR you could further improve. See https://github.com/Radiomics/pyradiomics/pull/197

PR #196, opened 2017-02-14
Add python3 support

jcfr:
This set of changes supersedes #195.

It includes changes required to run pyradiomics on Python 3. To gracefully handles the different between py2 and py3, it introduces the dependency on `six`. See https://pythonhosted.org/six/

Notes:
* instead of switching to using `dict.items()` (which is less efficient in py2), it uses [six.iteritems](https://pythonhosted.org/six/#six.iteritems)
* since MacOSX SimpleITK python wheels are not (yet) available for Python 3.x, corresponding TravisCI builds haven't been enabled.


.circleci-matrix.yml
```diff
@@ -3,12 +3,12 @@ env:
...
   - MANYLINUX_IMAGE=manylinux-x64 MANYLINUX_PYTHON=cp27-cp27mu
   - MANYLINUX_IMAGE=manylinux-x86 MANYLINUX_PYTHON=cp27-cp27mu
-#  - MANYLINUX_IMAGE=manylinux-x64 MANYLINUX_PYTHON=cp33-cp33m
-#  - MANYLINUX_IMAGE=manylinux-x86 MANYLINUX_PYTHON=cp33-cp33m
-#  - MANYLINUX_IMAGE=manylinux-x64 MANYLINUX_PYTHON=cp34-cp34m
-#  - MANYLINUX_IMAGE=manylinux-x86 MANYLINUX_PYTHON=cp34-cp34m
-#  - MANYLINUX_IMAGE=manylinux-x64 MANYLINUX_PYTHON=cp35-cp35m
-#  - MANYLINUX_IMAGE=manylinux-x86 MANYLINUX_PYTHON=cp35-cp35m
+  - MANYLINUX_IMAGE=manylinux-x64 MANYLINUX_PYTHON=cp33-cp33m
+  - MANYLINUX_IMAGE=manylinux-x86 MANYLINUX_PYTHON=cp33-cp33m
```
JoostJM: These two lines cause CircleCI to fail, as it tries to install numpy == 1.12.0, which does not support python 3.3 anymore.

jcfr: Indeed, I am now looking into adding few [env. markers](https://www.python.org/dev/peps/pep-0496/)

jcfr: Thinking more about it, as you suggested let's just remove support for Python 3.3. It is now worth the complication

.travis.yml
```diff
@@ -19,11 +19,6 @@ matrix:
 #      env:
 #        - PYTHON_VERSION=3.4.5
```
JoostJM: Doesn't this mean that python 3.x testing for Mac is still disabled?

jcfr: > python 3.x testing for Mac is still disabled?

Indeed. Waiting for SimpleITK wheels, we do not have a choice. See https://itk.org/SimpleITKDoxygen/html/PyDownloadPage.html

JoostJM:
@jcfr, last build of CircleCI failed, python 3.x couldn't find `venv` in some commandline argument, I suspect this has something to do with the virtual environment. Can you look into it?

jcfr:
> python 3.x couldn't find venv in some commandline argument, I suspect this has something to do with the virtual environment. Can you look into it?

I noticed. I will be looking at this later today.

jcfr:
> last build of CircleCI failed, python 3.x couldn't find venv in some commandline argument, I suspect 
> this has something to do with the virtual environment. Can you look into it?

I identified the problem and submitted a fix to upstream setuptools. See https://github.com/pypa/setuptools/pull/971

In the mean time, I will add a workaround.

jcfr:
Et voila. As soon as, Travis returns green. Will be integrating :fireworks:

.travis.yml
kmader: There are however conda packages for SimpleITK, would it make sense to get the mac tests working on miniconda/anaconda (https://anaconda.org/simpleitk/simpleitk) instead of standard python?

jcfr: That would make sense, the idea would be to submit a recipe to conda-forge. With that in place, the package would be available on conda on linux, macosx and windows. 

That said, conda-forge recipe are updated only when there is a release of the package. If you would like to setup continuous integration using miniconda, you would have to set it up on your own.

I currently do not have the bandwidth to set miniconda CI up for pyradiomics, in the mean time I would be happy to review PR.

@blowekamp Do we know when macosx wheels will be available for python 35 ?

blowekamp: I was just working on some scripts for the 1.0 release... I'll see if I can run them on the release branch too like the recent x86_64 0.10.0.post3 wheel.

PR #197, opened 2017-02-14
STYL: testUtils: Restore header sorting in test output

jcfr:
This commit partially reverts commit d84c793 (STYL: testUtils: Improve
test outputs) and restore the sorting of columns written into the
CSV file.

Considering that data are organized using "featureClassName + featureName"
instead of just "featureName", information will remain consistently organized.

Suggested-by: Joost van Griethuysen <j.v.griethuysen@nki.nl>

### DIFF
--- a/tests/testUtils.py
+++ b/tests/testUtils.py
@@ -76,6 +76,7 @@ class RadiomicsTestUtils:
     self._featureClassName = None
 
     self._testCase = None
+    self._testedSet = set()
 
     self._results = {}
     self._diffs = {}
@@ -98,12 +99,10 @@ class RadiomicsTestUtils:
     if self._featureClassName == className and self._testCase == testCase:
       return False
 
-    assert className in self.getFeatureClasses()
-    assert testCase in self.getTestCases()
-
     # First set featureClass if necessary, because if settings have changed, testCase needs te be reloaded
     if self._featureClassName != className:
       self._logger.debug('Setting feature class name to %s', className)
+      assert className in self.getFeatureClasses()
 
       self._featureClassName = className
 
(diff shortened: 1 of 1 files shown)


JoostJM:
@jcfr, I'd like to change this a bit, as the current version is a bit buggy when you only test certain cases...

tests/testUtils.py
```diff
@@ -245,18 +247,23 @@ def writeCSV(self, data, fileName):
...
+      csvFileWriter.writerow(header)
+      for testCase in testedCases:
+        thisCase = data[testCase]
+        thisCase['testCase'] = testCase
+        row = []
+        for h in header:
+          row = row + [thisCase.get(h, "N/A")]
+        csvFileWriter.writerow(row)
+      csvFile.close()
+      self._logger.info('Wrote to file %s', fileName)
```
jcfr: Should it be `self._logger.info('Wrote to file %s' % fileName)` ?

tests/testUtils.py
```diff
@@ -245,18 +247,23 @@ def writeCSV(self, data, fileName):
...
-      row = []
-      for h in header:
-        row = row + [thisCase.get(h, "N/A")]
-      csvFileWriter.writerow(row)
-    csvFile.close()
-    self._logger.info('Wrote to file %s', fileName)
+    # Get the headers from the first testCase in _testedSet
+    # If no tests were run, the length of _testedSet will be 0, and no files should be written
+    if len(self._testedSet) > 0:
+      csvFile = open(fileName, 'w')
```
jcfr: You could use:

```
with open(fileName, 'w') as csvFile:
  csvFileWriter = csv.writer(csvFile, lineterminator='\n')
  ...
```

tests/testUtils.py
JoostJM: Not necessarily. As it is a logging message, a format is applied by default. So both would work.

JoostJM: https://docs.python.org/2/library/logging.html?highlight=logger.debug#logging.Logger.debug

tests/testUtils.py
```diff
@@ -246,18 +248,23 @@ def writeCSV(self, data, fileName):
...
+        header = ['testCase'] + header
+        csvFileWriter.writerow(header)
+        for testCase in testedCases:
+          thisCase = data[testCase]
+          thisCase['testCase'] = testCase
+          row = []
+          for h in header:
+            row = row + [thisCase.get(h, "N/A")]
+          csvFileWriter.writerow(row)
+        csvFile.close()
```
jcfr: Close is not required anymove since we use `with`

JoostJM: Thanks for pointing that out, I fixed it in my latest commit.

PR #200, opened 2017-02-15
Support for faster C-based matrix and shape surface computation

jcfr:
This topic is "equivalent" to #158 but was rebased on the current master on 2017-02-15

Note that it doesn't (yet) include the following:
* [x] Update of C extensions module to support Python 3
* [ ] Integration of [scikit-build](http://scikit-build.readthedocs.io/) to support seamless building on multiple platform


radiomics/shape.py
```diff
@@ -50,6 +50,7 @@ def __init__(self, inputImage, inputMask, **kwargs):
       self.SurfaceArea = self._calculateSurfaceArea()
 
   def _calculateSurfaceArea(self):
+    self.maskArray = self.maskArray.astype('int')  # needed for the interpolate function to work correctly
```
JoostJM: @jcfr, Why did you push this commit? If you look at L42, you can see that the datatype is already forced to `int`...

JoostJM:
@Radiomics/developers, what is the preference, as this concerns a branch which was created before the repository was made public. I was working in parallel to @jcfr to update the C extensions for integration with the much changed master. However, I opted to make a clean branch from the current master, cherry picking and making new commits. This is a somewhat more 'clean' branch, with the old bugfixes quashed into the first commits, and without first added, then deleting the C extension for the experimental feature classes. It is located [here](https://github.com/JoostJM/pyradiomics/tree/c-matrices_revised).

JoostJM:
@jcfr, Do you want to combine the integration of scikit-build with the C extensions in this PR, or do you want to address that in a different PR? Seeing as the tests are currently passing, I think it may be more logical to address this in a different PR.

jcfr:
> Do you want to combine the integration of scikit-build with the C extensions in this PR, or do you want 
> to address that in a different PR? Seeing as the tests are currently passing, I think it may be more 
> logical to address this in a different PR.

A different PR will make more sens

>  more 'clean' branch, with the old bugfixes quashed into the first commits, and without first adding, 
> then deleting the C extension for the experimental feature classes.

Just because, here is an updated branch without adding/deleting the feature. It now contains only three commits. ( Initially, I just thought it would be nice to have the "not ready for production" implementation of the other features still available in the history.)
See https://github.com/Radiomics/pyradiomics/compare/c-matrices-rebased-jcfr-20170216?expand=1

May be we could keep part of this commit message: https://github.com/Radiomics/pyradiomics/commit/6461197a4e9270aefaa8e93182aac3fe0f28ed16

All of that said, it makes sense to move forward with your topic, it is cleaner. You will find below an updated version where I
* squashed the style and compilation fixes where they belong
* remove the dependency on cython (it is not used as far as i can tell ... no pyx files)

See https://github.com/Radiomics/pyradiomics/pull/202

Comparing the two topic (`c-matrices-rebased-jcfr-20170216 [jcfr]` and `c-matrices_revised [JoostJM]`), the only major differences is in that in your topic, the weighting in glcm.py is removed. What this intentional ? 

This link illustrate the difference: https://github.com/Radiomics/pyradiomics/compare/c-matrices-rebased-jcfr-20170216...c-matrices-rebased-jcfr-20170216-diff-joost-revised#diff-c8aee871337c890a9b744f600b6e8d67

JoostJM:
>Comparing the two topic (c-matrices-rebased-jcfr-20170216 [jcfr] and c-matrices_revised [JoostJM]), the only major differences is in that in your topic, the weighting in glcm.py is removed. What this intentional 

It is not removed as such. Weighting and removal of empty angles is done in python for both the python calculated and C calculated matrices. To prevent bugs, I put this functionality in a separate function (`_applyMatrixOptions`) and subsequently call that from the `_calculateMatrix` and `_calculateCMatrix` functions.

There are some other differences as well, I'll list them here:
 - Standardized the naming of matrix calculation functions (`_calculateMatrix` and `_calculateCMatrix`). This is needed to enable a simplified testing script for `test_cmatrices`.
 - In C calculation of GLSZM, I used a temporary array to store the calculated zones, with zones defined as pairs of two elements, with the first item defining the gray level, and the second the size. This is much less memory intensive (only 2 * NumVoxels + 1: this size is needed when ROI consists of only zones of size 1). This matrix is then loaded into the output GLSZM, which can be initialized directly in its correct size (instead of NumGrayLevels * NumVoxels, which is required otherwise). Before, I created this temporary array as a python numpy array object, and getting a pointer to the data. Here, I directly allocate memory for the array using `calloc` and release it with `free`, as the temporary data only exists in C.
- I renamed the function to force full python to what I think is a more logical name, and added some code to prevent an import error in the featureclasses if the C extensions don't work (`cMatrices` and `cShape` are imported into the featureclass, even if not loaded, so if they are not loaded, they must still be defined, I achieve this by assigning `None` if the import fails.

Finally, I want to parameterize whether full-python mode should be forced, so it can be incorporated into a parameterfile, or passed in `kwargs` to `featureextractor`. I will push this later.

JoostJM:
> May be we could keep part of this commit message: 6461197

I added an updated version of this commit message to the discussion thread in #202. Furthermore, I think we should incorporate this in some form in the release notes when we declare the next stable release that has this integrated.

jcfr:
> To prevent bugs, I put this functionality in a separate function 

Well done. Look at it again, makes sense.

> some other differences as well, I'll list them here

Thanks for the details explanation :+1:

jcfr:
Closing this PR. It is superseded by #202

PR #202, opened 2017-02-16
C matrices revised jcfr 20170216

jcfr:


JoostJM:
This topic is an alternative to #200 and supersedes #158.

JoostJM:
Stats (Profiling 5 testcases):
  GLCM 6913 ms -> 3 ms
  GLSZM 12064 ms -> 58 ms
  GLRLM 1850 ms -> 10 ms
  Surface Area 3241 -> 1 ms

This PR adds two C Python extensions and associated tests:

* _cmatrices: C implementation for matrix computation associated with GLCM, GLSZM. and GLRLM features

* _cshape: C implementation for Shape surface computation

* tests/test_cmatrices: testing for matrix equality: Tests whether the python generated matrix is equal to the matrix generated by the C implementations (allows for machine precision errors). Also compares the python and C generated surface computations similarly.

Details
-------

* Add docstring to C modules

* Use of C implementation optional. At initialization, the package tries to use C, but if loading has failed, or calculation is forced to python, python is used. Note that the import of _cmatrices is done
  after initialization of logger. This ensures error will be logged if import fails.

* GLSZM: Use "char" datatype for mask. (It is signed char in GLSZM).

* C code is consistent with C89 convention. All variables (pointers for python objects) are initialized at top of each block.

* GLSZM: Use `calloc` and `free` for the temporary array holding the calculated zones.

Optimizations
-------------

* GLSZM:

 - Define temporary array for holding the calculated zones. During calculation, the matrix must be able to store all possible zones, ranging from zone size 1 to total number of voxels (Ns), for each gray level (Ng). In this case, the GLSZM would be initialized with size Ng * Ns, which is very memory intensive. Instead, use a temporary array of size (Ns * 2) + 1, which stores all calculated zones in pairs of 2  elements: the first element holds the gray level, the second the size of the calculated zone. The first element after the last zone is set to -1 to serve as a stop sign for the second function, which translates the temporary array into the final GLSZM, which can be directly initialized at optimum size.

  - Uses "while" loops. This allows to reduce the memory usage. We observed that with recursive functions it was 'unexpectedly' failing.

  - Optimized search that finds a new index to process in the region growing.

jcfr:
And at some point, we should probably look into running the C code through tool like `valgrind --memcheck`.

jcfr:
Otherwise, LGTM

JoostJM:
@jcfr Here are two stack overflow threads covering using valgrind for python C extensions checking:

- http://stackoverflow.com/questions/3982036/how-can-i-use-valgrind-with-python-c-extensions
- http://stackoverflow.com/questions/20112989/how-to-use-valgrind-with-python

PR #207, opened 2017-02-20
Update readme

JoostJM:
This PR updates the documentation to better reflect the current master.

Additionally, there were some bugs left over from the added compatibility for Python 3 and the changed implementation of the image filters. These were all contained in the additional scripts in the `bin` folder and therefore did not affect the continuous testing.

Finnally, @Radiomics/developers, do we want to increment the version number again? The additional python 3 compatibility and integration of C extensions are a fairly large change.


fedorov:
I agree, definitely the version should be updated, perhaps even the minor version, not the patch version.

JoostJM:
> I agree, definitely the version should be updated, perhaps even the minor version, not the patch version.

I agree to update the minor version.

fedorov:
Should it be 1.1.0 instead of 1.1?

PR #209, opened 2017-02-22
Implement normalization function in imageoperations

JoostJM:
Optionally normalize gray value intensities before any resampling or filters are applied. Image is normalized on all gray values (not just those in the ROI). Disabled by default. Controlled by 3 new params (`normalize`, `normalizeScale` and `removeOutliers`).

### DIFF
--- a/data/paramSchema.yaml
+++ b/data/paramSchema.yaml
@@ -25,7 +25,9 @@ mapping:
         range:
           min-ex: 0
       removeOutliers:
-        type: bool
+        type: float
+        range:
+          min-ex: 0
       resampledPixelSpacing:
         seq:
           - type: int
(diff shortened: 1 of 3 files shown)


JoostJM:
As discussed today, slightly changed the implementation of the normalize function:
Removal of the outliers applied after normalization, before applying the scale. Additionally, outlier threshold can now be customized, as parameter `removeOutliers` is a float specifying this threshold. If this parameter is omitted from the parameter file or `kwargs` passed at initialization of `featureextractor`, it is set to `None`, and no outliers are removed. This is the default behaviour. Setting it to `None` is not possible in the parameter file, this throws an error (i.e. if no outliers should be removed, omit the parameter from the parameter file).

PR #210, opened 2017-02-23
Fix pyradiomicsbatch error with python 3

JoaoSantinha:
In python 3.x (tested 3.4) featureVector.keys() and
featureVector.values() don’t return a sequence expected in
writer.writerow() causing the error:

writer.writerow(featureVector.keys())
_csv.Error: sequence expected

writer.writerow(featureVector.values())
_csv.Error: sequence expected


JoaoSantinha:
@Radiomics/developers

JoostJM:
@JoaoSantinha, good catch! Thanks.

PR #212, opened 2017-02-24
setup: use scikit-build for building cmatrices and cshape C extensions

jcfr:
This commit introduces a new dependencies to "scikit-built" and "cmake".

scikit-build is a drop-in replacement to setuptools.setup function
allowing to easily compile and package extensions (C/C++/Cython) by
bridging CMake and setuptools. See http://scikit-build.org

CMake is is an open-source, cross-platform family of tools designed
to build, test and package software. See https://cmake.org

Currently, scikit-build and cmake have to be explicitly (see [1]) installed
on the system. This could be done by simply doing:

  pip install scikit-build cmake

How does it work ?
------------------

In addition to simplifying setup.py, two new file have been added:

```
<root>
  |
  |---- CMakeLists.txt                          (1)
  .
  .
  .
  |--- ...src
  |        |--- CMakeLists.txt                  (2)
  .
  .
```

The first CMakeLists.txt specifies requirements for

* Python interpreter,

* the associated python libraries

* and a "PythonExtension" modules containing convenience CMake
  functions allowing to easily build extension by encapsulating
  system introspection and platform specific logic.

and include the subdirectory associated with the other CMakeLists.txt.

The second CMakeLists.txt is specific to "cmatrices" and "cshape"
extensions and it specifies:

* requirement for NumPy

Then, it declares:

* libraries for each extension, associate python extension specific properties

* and finally add an install rule specifying where in the package
  hierarchy the extension should be installed.

Et voila,

[1] Note that improvement to python packaging will be available shortly and
it will be possible to include scikit-build and cmake directly in pyproject.toml
See here for more details: https://www.python.org/dev/peps/pep-0518/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,7 @@ docs/_build/
 
 # PyBuilder
 target/
+
+# scikit-build
+_skbuild
+MANIFEST
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,10 @@
+cmake_minimum_required(VERSION 3.7)
+
+project(radiomics)
+
+find_package(PythonInterp REQUIRED)
+find_package(PythonLibs REQUIRED)
+find_package(PythonExtensions REQUIRED)
+
+add_subdirectory(radiomics/src)
+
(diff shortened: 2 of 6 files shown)


fedorov:
> This commit introduces a new dependencies to "scikit-built" and "cmake".

What are the consequences of adding scikit-build dependency for the SlicerRadiomics extension integration into Slicer?

jcfr:
Closing. Superseded by #311

PR #213, opened 2017-03-02
ENH: check if feature class name is valid

fedorov:


radiomics/featureextractor.py
```diff
@@ -284,6 +284,9 @@ def enableFeatureClassByName(self, featureClass, enabled=True):
     """
     Enable or disable all features in given class.
     """
+    if not featureClass in self.getFeatureClassNames():
```
JoostJM: @fedorov, flake 8 fails due to this line: if you compare with "not" and "in", flake 8 requires you to put these together, i.e. `if featureClass not in self.getFeatureClassNames():`

fedorov: Thanks! I admit I forgot to follow up and check the tests after submitting the PR.

fedorov:
@JoostJM the checks are now passing.

PR #215, opened 2017-03-06
Implement forced 2D texture extraction.

JoostJM:
Add additional parameters to optionally force texture calculation to 2D. This is enabled by setting parameter `force2Dextraction` to True. Then, all angles moving across 'slices' are removed. The dimension which specifies the slice is defined in `force2Ddimension`. For example, if enabled and dimension is 0, texture is extracted only in the xy-plane (with dimension 0 specifying the 'z' dimension as slice-dimension). Forced 2D extraction is switched off by default.

Additionally, enable specifying custom distances between center voxel and neighbors. These distances are specified in the `distances` parameter as a list of integers. By default `disances = [1]`.

### DIFF
--- a/radiomics/imageoperations.py
+++ b/radiomics/imageoperations.py
@@ -76,7 +76,7 @@ def generateAngles(size, **kwargs):
 
     - distances [[1]]: List of integers. This specifies the distances between the center voxel and the neighbor, for
       which angles should be generated.
-    - force2Dextraction [False]: Boolean, set to true to force a by slice texture calculation. Dimension that identifies
+    - force2D [False]: Boolean, set to true to force a by slice texture calculation. Dimension that identifies
       the 'slice' can be defined in ``force2Ddimension``. If input ROI is already a 2D ROI, features are automatically
       extracted in 2D.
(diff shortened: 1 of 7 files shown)


radiomics/imageoperations.py
```diff
@@ -51,33 +51,63 @@ def binImage(binwidth, parameterMatrix, parameterMatrixCoordinates):
...
+      identifies the 'z' dimension (axial plane feature extraction), and features will be extracted from the xy plane.
+      Similarly, 1 identifies the y dimension (coronal plane) and 2 the x dimension (saggital plane). if
+      ``force2Dextraction`` is set to False, this parameter has no effect.
+
   :return: numpy array with shape (N, 3), where N is the number of unique angles
   """
 
+  distances = kwargs.get('distances', [1])
+  force2Dextraction = kwargs.get('force2Dextraction', False)
+  force2Ddimension = kwargs.get('force2Ddimension', 0)
```
fedorov: Minor comment - the names of these two are a bit confusing, since to me they both imply boolean type. Maybe call them `force2D` and `force2Ddimension`?

JoostJM: @fedorov good suggestion, thanks! Will change.

PR #218, opened 2017-03-08
Revise Compactness 1 Formulas

JoostJM:
Flip exponent (2/3 -> 3/2) of Compactness 1 feature formula. The previous exponent was most likely an error, as this resulted in a formula that was not dimensionless, as the description indicates. The formula with the new exponent is dimensionless.

Compactness1, Compactness2, Sphericity and Spherical disproportion are mathematically correlated. Update the documentation to reflect this (provide equations specifying the correlation between the different features). Also disable Compactness 1 and Compactness 2 in the default/example parameter file in the bin folder to emphasize the redundancy of these features.

Updated the formulas and documentation thereof to have simplified formulas where possible.


pieper:
LGTM - sounds like a good fix.

philippguevorguian:
> Flip exponent (2/3 -> 3/2) of Compactness 1 feature formula. The previous exponent was most likely an error, as this resulted in a formula that was not dimensionless, as the description indicates. The formula with the new exponent is dimensionless.
> 
> Compactness1, Compactness2, Sphericity and Spherical disproportion are mathematically correlated. Update the documentation to reflect this (provide equations specifying the correlation between the different features). Also disable Compactness 1 and Compactness 2 in the default/example parameter file in the bin folder to emphasize the redundancy of these features.
> 
> Updated the formulas and documentation thereof to have simplified formulas where possible.

Hi @JoostJM, 

I'm a radiomics researcher attempting to validate a model proposed by your associate Hugo Aerts in his landmark paper "[Decoding tumour phenotype by noninvasive imaging using a quantitative radiomics approach](https://www.nature.com/articles/ncomms5006)". I'm reaching out to you here because I have a question about the calculation of compactness1 as it pertains to this work and the validation paper "[External validation of a prognostic CT-based radiomic signature in oropharyngeal squamous cell carcinoma](https://www.tandfonline.com/doi/full/10.3109/0284186X.2015.1061214)". 

In both papers, the [supplementary materials](https://github.com/AIM-Harvard/pyradiomics/files/5755985/Supplementary.pdf) describe compactness1 as having been calculated with an exponent of 2/3 on the area. I understand the value of changing the exponent to 3/2 such that the metric is dimensionless as its description states. However, I am curious to know how this feature was calculated in these papers for the purposes of a successful validation. Specifically, I want to know if they used an exponent of 2/3 as reported in the supplementary materials and this was then changed in this update of pyradiomics **or** if this was a typo that was then implemented in pyradiomics and needed changing afterwards. Any guidance would be appreciated.

Thanks in advance

PR #219, opened 2017-03-10
Change binning

JoostJM:
This PR changes the way the images are binned. This does not affect the values in the test cases, it only affects the binning for specific cases:

When the maximum gray level inside the ROI is equally dividable, it will be equal to the rightmost bin edge.
`numpy.histogram` includes these values inside the topmost bin (i.e. the rightmost bin is closed, whereas the others are half-open). However, in `numpy.digitize`, the rightmoste 'bin' is also half open, and these values will then be set to the number of bins + 1. For image discretization this is better if you consider the following example:

range 5 (unique values 1, 2, 3, 4, 5), binWidth 1: bin edges: (1, 2, 3, 4, 5)

This PR will do binning in such a way, that the values are not changed. Before this PR, voxels with value 5 would be set to 4.

Additionally, set the default value for `voxelArrayShift` to 0, to prevent users from accidentally using the fairly large array shift which is the current default. This PR also changes the documentation to make users more aware of this, with updates on the definition of the parameter, and incorporation of the parameter in the feature formulas that are affected by this parameter.


pieper:
I don't completely understand your commit message - I think you are saying that you are changing the behavior to be consistent with numpy.histogram, is that correct?  If so LGTM.

radiomics/imageoperations.py
```diff
@@ -13,42 +13,64 @@
 logger = logging.getLogger(__name__)
 
 
-def getHistogram(binwidth, parameterValues):
-  """
+def getBinEdges(binwidth, parameterValues):
```
fedorov: I just now realized that numpy has its own set of strategies for calculating bins. I think it would be helpful to give user an option to access those numpy strategies. I don't think this is possible right now.

Related question - should we raise the question of bin calculation withe radiomics standardization effort group? It does not look like their document discusses this at the moment.

JoostJM:
> I don't completely understand your commit message - I think you are saying that you are changing the behavior to be consistent with numpy.histogram, is that correct? If so LGTM.

In fact, the opposite. `numpy.histogram` has a closed interval for the rightmost bin (i.e. if you have gray values in range 1 - 5, with bin width 1, numpy.histogram will assign values 4 and 5 to bin 4, whereas the lower gray values are assigned to their own bin: 1 -> 1, 2 -> 2, 3 -> 3). `numpy.digitize` does not do this (i.e. the gray level 5 would be assigned to len(bins) + 1 -> 4 + 1 = 5). 

The reason I removed the line of code which made the `numpy.digitize` behaviour consistent with `numpy.histogram` is that it prevents 'switching off' binning by specifying binWidth = 1.

In it's current form, the behavior of `numpy.histogram` and `numpy.digitize` differs for the voxels with max(gray level), where max(gray level) mod 0 = 0 (so it's just in very specific cases). An alternative would be to change the behavior of `numpy.histogram` to match `numpy.digitize` (i.e. make sure that gray level 5 would be binned to a 5th bin in the example above).

pieper:
Okay, that's clearer now.

radiomics/firstorder.py
```diff
@@ -5,27 +5,31 @@
...
   :math:`p(i)` be the normalized first order histogram and equal to :math:`\frac{\textbf{P}(i)}{\sum{\textbf{P}(i)}}`
 
-  Following addiotional settings are possible:
+  Following additional settings are possible:
 
-  - voxelArrayShift [2000]: This amount is added to the gray level intensity in Energy, Total Energy and RMS, this is to prevent negative values from occuring when using CT data.
+  - voxelArrayShift [0]: Integer, This amount is added to the gray level intensity in features Energy, Total Energy and
+    RMS, this is to prevent negative values. __If using CT data, or data normalized with mean 0, consider setting this
+    parameter to a fixed value that ensures non-negative numbers in the image. Bear in mind however, that the larger the
+    value, the larger the volume confounding effect will be.__
```
JoostJM: I added the suggestion at the definition of `voxelArrayShift` parameter in the documentation (at the start of firstorder feature class)

radiomics/imageoperations.py
JoostJM: > I just now realized that numpy has its own set of strategies for calculating bins. I think it would be helpful to give user an option to access those numpy strategies. I don't think this is possible right now.

It's a good question. We discussed this in one of the hangouts some months ago (where I proposed adding the possibility for a fixed bin count). We decided then not to incorporate it to prevent confusion. Additionally, do the other options yield equally spaced bins? This is required for `numpy.digitize`. If not, we have to decide whether or not to enable this for just histogram related features (Entropy and Uniformity in firstorder)

> Related question - should we raise the question of bin calculation withe radiomics standardization effort group? It does not look like their document discusses this at the moment.

They already took it up, it's in their latest workversion (but not available online yet I think).

They already took it up, it's in their latest workversion

fedorov: > do the other options yield equally spaced bins?

I didn't test, but looking at the code, it seems that the binning strategy just calculates the width of the bin: https://github.com/numpy/numpy/blob/v1.12.0/numpy/lib/function_base.py#L695

JoostJM: > I didn't test, but looking at the code, it seems that the binning strategy just calculates the width of the bin: 

It looks like that to me too. Still raises the question on whether to enable this. Main issue remains that any of these functions calculate bin widths on the input range, meaning that the comparability of gray values may be limited...

PR #220, opened 2017-03-14
Update PyRadiomics Logging

JoostJM:
Update logging, with mainly added debug messages on various functions throughout the toolbox.

Additionally, change how logging is set up in the various scripts used to run PyRadiomics. Instead of using a root logger, keep `radiomics` logger as 'root' logger, with script/batch specific log messages in the `radiomics.script` and `radiomics.batch` logger. This excludes the PyKwalify debug messages from the PyRadiomics log when level is set to DEBUG.


JoostJM:
@Radiomics/developers. As discussed in the last hangout, I changed the progress reporting in PyRadiomics to make use of the logging package where possible. This therefore also removes the `verbose` setting. The only output not directly handled by the logging is the progress bar in GLRLM and GLCM. However, the output of the progress bar is also to `stderr`, and it is only enabled when python calculation is enabled and the logging level of the handler (to `stderr`) has level INFO or DEBUG.

Issue #18, opened 2015-12-23
Add checks to ensure image and label correspondence 


vnarayan13:
Add checks (in base class?):
1) image and label have similar (within tolerance) base pixel spacing
2) image and label have similar origins (does this vary with dimension?)
3) image and label have similar 3-space orientation (space directions in nrrd header?)
4) image and label have similar dimensions (not necessary, but need to somehow check registration)

JoostJM:
In #223, I added some checks (including thoses suggested here). The checks described here are applied by SimpleITK when the bounding box is calculated using the `LabelStatisticsImageFilter`, with tolerance specified in the `ProcessObject`. 

In addition to these checks, also checks some size constraints (can't extract features from a single voxel).

2 points with these checks:

1. checks will fail if ROI is a subvolume of original image. I left it like this, because these should not be encouraged, will represent (hopefully) a minority of cases and to write a check for this will most likely not be very performance efficient (as it requires you to compute min and max bound for both image and mask in realworld coordinates and then check validity)
2. these checks are applied after resampling, as the size constraints are important for the ROI after resampling (otherwise, there still could be a single voxel extraction). This does mean that if there is a mask and image that does not match up before resampling, no error will be thrown, as resampling forces them to the same geometry.

JoostJM:
What would be possible is to use the same `LabelStatisticsImageFilter` during resampling (also requires the bounding box). This would do the geometry check also before resampling...

JoostJM:
resolved by #223

PR #223, opened 2017-03-22
Add checks for mask size and dimensions.

JoostJM:
Add function to check if ROI is large enough for computation. 
First checks if label is present in the mask, then checks if the number of dimensions in which the size > 1 is at least the minimum number of dimensions (default: 1). Finally an an optional check for a minimum number of voxels in the tumor region can be performed (switched off by default).

This check is performed in `featureextractor` after the loading of the image and any image pre-processing steps (e.g. resampling), but before features are extracted or filters applied.

Addresses #18 
cc @Radiomics/developers


bin/addClassToBaseline.py
```diff
@@ -74,7 +74,8 @@ def main():
 
     provenance = extractor.getProvenance(imagePath, maskPath, mask)
 
-    image, mask, bb = imageoperations.cropToTumorMask(image, mask)
```
fedorov: I would avoid using "Tumor". Maybe "cropToMaskBoundingBox", or "cropToMaskROI".

fedorov: Sorry, I understand this is not related to the essence of this specific PR

fedorov:
LGTM otherwise!

bin/addClassToBaseline.py
JoostJM: > Sorry, I understand this is not related to the essence of this specific PR

No, it's not, but it's still a good suggestion. I am running over the documentation for errors, will update this then also.

Issue #189, opened 2017-02-09
Build around Pandas / DataFrames instead of CSVWriter


kmader:
The library would be a lot more flexible if it were built around Pandas and DataFrames since they are better suited for looking at collections of data. They also allow for many more export options like json, html, csv, and have better support for types inside columns extendable to figures and arrays

![image](https://cloud.githubusercontent.com/assets/116120/22780515/4c6e8d6c-eebf-11e6-955a-50ef0499f565.png)

JoostJM:
@kmader, thanks for the suggestion! We've discussed this with the team and we'll start looking how to best implement this into PyRadiomics.

JoostJM:
@kmader, I've added an example of how you can use pandas to process the result of PyRadiomics in PR #228 . It is not needed for data handling inside PyRadiomics and the returned result from the feature extractor can easily be converted to a pandas Series, thereby removing the need to rewrite a lot of source code and have pandas as an additional dependency.

In PR #225 pandas is used much more extensively throughout the PyRadiomics package, but this does not change output or performance, but will need pandas as another dependency, so this PR will most likely be closed and not merged.

PR #225, opened 2017-03-24
Add pandas

JoostJM:
Add use of pandas to handle data in PyRadiomics

Affects all data handling in the feature base class and feature extractor, which now passes the calculated features as a pandas Series with the feature names as labels.
Change testing to load baseline as a pandas dataframe and also process/check/store results in a pandas DataFrame.

Update both scripts and `batchprocessing.py` to use pandas. Results are now stored in a pandas DataFrame and only written to disk upon completion of the extraction. This ensures that all calculated features are written, and writes "NaN" for features that failed or produced a NaN value.

Update documentation to reflect different methods for data handling.


JoostJM:
Addresses #189

JoostJM:
Even though returned type after extraction is now a pandas Series, no recoding of scripts is required, as the Series can be used similarly to a dictionary. For this, I tested `six.iteritems` and `keys()`, which worked as expected. I did change the batch scripts to better profit from the additional functionality of pandas though.

fedorov:
I think @blezek has a lot of experience with pandas - would be great if he could review this one!

JoostJM:
The changes in this PR up to now are to switch data handling to pandas, without extending on the functionality. I did not add additional functionality, such as storing in HDF5, because I first want to make sure the current functionality of PyRadiomics is preserved, before adding new things.

blezek:
Changes look good.  My only concern is the list of requirements for pyradiomics is getting longer.  Since Pandas isn't critical for processing, it may make sense to move it to a `util` module, much like `vtk` has the `numpy` conversion in a optional module.

JoostJM:
@blezek, I agree it is yet another dependency. This also made me decide not to implement the hdf5 support, as this also need an additional dependency.
By putting pandas in a separate `util` module, do you mean that inside pyradiomics it is kept as dictionaries and only converted to pandas if requested by the user?

blezek:
@JoostJM Yes, I think that would be nice, or if you saved the final conversion until right at output time, it could be an optional output,i.e. `output_as_pandas=True`

I really don't have strong feelings on this one way or another, just hate to see Pandas being a requirement for a nice to have, but minor feature.

pieper:
I agree with Dan - my impression is that the conversion to pandas should be
pretty straightforward as a last step so it's not really a dependency we
need.  It would be different if we had a reason to use pandas internally to
represent or work with the feature values themselves.

JoostJM:
@pieper, in this PR I did use pandas more extensively, mainly to simplify the reading / storing and use of baseline when testing the features. Although this is somewhat simpler, and better guarantees everything is written out correctly, it is not necessary to run the tests, and if tests pass, output is also identical.

Therefore, I made a revised PR, where pandas is just an optional feature.

JoostJM:
Superseded by #228

PR #228, opened 2017-03-28
Add pandas revised

JoostJM:
Supersedes #225 

Pandas is added in this PR as an optional feature, and results are only converted to pandas Series just prior to return to user. To illustrate use of pandas in PyRadiomics, a new batch example is added and the helloRadiomics script and notebook is updated with the additional call to enable pandas (commented out by default).

This PR also changes the way general info is stored slightly (retaining data types). This mainly affects how the result is stored when using JSON format, where the settings will also be stored as a JSON formatted dictionary string (i.e. a nested dictionary), instead of just as a string.

Finally, change the format of the input CSV for the batch command line script and batch example scripts. This CSV should now contain headers and allows a variable number of columns (which are all copied to the output). Columns must contain 'Image' and 'Mask', indicating the columns with the file locations of the image and mask, respectively.


JoostJM:
I removed pandas use from the source code entirely and instead provided an example of the batchprocessing script instead.

This is because to add pandas as an optional feature in PyRadiomics involves a lot of code to check whether pandas can and should be used, whereas the actual use is just a simple conversion just before the result is returned.

This conversion is done in this new example (which assumes the user has pandas installed), with additional code showing how to combine results (pandas Series) into a pandas DataFrame and how to use pandas for reading / writing csv.

Issue #230, opened 2017-03-29
Remove non-core dependencies


JoostJM:
To keep number of requirements down, remove dependencies that are not needed for (regular) use of core functions.

*  [x] Move nose dependencies to requirements-dev
*  [ ] Move Sphinx dependencies to requirements-docs
*  [x] Remove dependency of tqdm

jcfr:
> Move Sphinx dependencies to requirements-docs

For now, I think having them in `requirements-dev.txt` is good enough. I may revisit things while working on the Slicer radiomics  integration.

jcfr:
> Remove dependency of tqdm

For progress reporting, I also like the idea of context manager. 
See http://pocco-click.readthedocs.io/en/latest/utils.html#showing-progress-bars and https://github.com/girder/girder/blob/834cf29acc3538b38ba24e6e7203852924110843/clients/python/girder_client/__init__.py#L823

JoostJM:
> For progress reporting, I also like the idea of context manager.

The problem is not so much that tqdm does not work as we expect, just that it is yet another dependency, but one that is nearly never used (only used in full-python mode for 2 of the 5 feeature classes)

JoostJM:
Removal of `tqdm` dependency addressed by #232

JoostJM:
> For progress reporting, I also like the idea of context manager

I looked at your code, really nice! I will see how to adapt this to PyRadiomics. In the mean time I opened a PR which now exposes a kind of quick-and-dirty fix.

JoostJM:
@jcfr, I took your example with a dummy progress reporter in girder and adapted it to PyRadiomics. I tested it with both `tqdm` (can be used directly) and `click` (used in a simple wrapper class). I also pass the iterable (gray levels) to the progress bar instead of a length and calls to `update`, because this was more similar between `tqdm` and `click`.

JoostJM:
Addressed by #231

PR #216, opened 2017-03-07
Update Documentation

JoostJM:
Features classes are documented in separate section ("Feature Definitions"). Also, add links to corresponding sections in the index file.
Finally, rename headers and order of documentation to better reflect PyRadiomics structure.

PR #231, opened 2017-03-29
requirements: Move nose and sphinx to dev requirements. See #230

jcfr:


requirements-dev.txt
```diff
@@ -1,2 +1,6 @@
+nose>=1.3.7
+nose-parameterized>=0.5.0
 flake8
 flake8-import-order
+sphinx>=1.4
+sphinx_rtd_theme>=0.2.2
```
JoostJM: would this require user to install the dev version of pyradiomics before documentation can be built? (or have users install `sphinx` and `sphinx-rtd-theme` manually?)

jcfr:
> install the dev version of pyradiomics

What do you mean by dev version ? 

To summarize, as it was before building the doc can be achieved doing:

```
git clone https://github.com/Radiomics/pyradiomic
cd pyradiomic
pip install -r requirements.txt  -r requirements-dev.txt
cd docs; make html
```

The advantage of this commit is that it avoids installing sphinx related packages while simply doing `pip install pyradiomics`.

I would also suggest to remove this line: https://github.com/Radiomics/pyradiomics/blob/d00212cd407d96604d5d9f1581661547efd3be3a/MANIFEST.in#L15

JoostJM:
> To summarize, as it was before building the doc can be achieved doing:

Building the docs does require pyradiomics to be installed, as it uses the __version__ of pyradiomics to specify the package version when building the docs.

This is not needed for the online documentation on readthedocs, but it is reflected in the downloadable pdf or custom build documentation (this was set at 0.0.1). I changed this in #216 to use the actual version of pyradiomics.

JoostJM:
> What do you mean by dev version ?

`python setup.py develop`? Or when are the development requirements installed by the setup script?

JoostJM:
> I would also suggest to remove this line: https://github.com/Radiomics/pyradiomics/blob/d00212cd407d96604d5d9f1581661547efd3be3a/MANIFEST.in#L15

I agree

JoostJM:
@jcfr, can you update the manifest and merge?

JoostJM:
@jcfr, what is the status on this PR?

JoostJM:
Rebased onto master to resolve conflict (due to the removal of the `tqdm` dependency in #232)

PR #232, opened 2017-03-29
Remove dependency for `tqdm`

JoostJM:
Remove dependency for the tqdm package, as this is rarely used (only for GLCM and GLSZM in full-python mode).
Instead provide 3 variables which can hold function definitions used for instantiating and interacting with a progress bar.

In the helloRadiomics script and notebook, this changed functionality is illustrated with a commented out section in the script and a separate cell in the notebook. This section and cell assume the `tqdm` package is installed and use it to show a progressbar during calculation of the glcm matrix in full-python mode.

TODO before merge:

 - [x] Document the requirements for the progressReporter specification in the developers documentation
 - [x] Add documentation to the `_getProgressReporter` function in `__init__.py`


radiomics/__init__.py
```diff
@@ -161,6 +161,18 @@ def getInputImageTypes():
   return _inputImages
 
 
+def getProgressFunctions():
```
jcfr: should that function be private: `_getProgressFunctions()` ? Or should the variable `(init|report|close)progress` be private global ?

Either way we need to document the signature of the functions too.

PR #233, opened 2017-04-05
Update documentation

JoostJM:
Change style of feature documentation to be more easily adaptable for a article supplement.

Update ReadMe files in `ROOT` and in `ROOT\data`.

Add additional FAQs.

Remove the SumVariance feature and rename SumVariance2 to SumVariance. This is due to the fact that the original SumVariance in PyRadiomics was based on a literal interpretation of the definition in the Haralick paper, where it is possible that the reference to entropy instead of average was a typo (as in f7, it referred to f8 instead of f6).


JoostJM:
N.B. This also includes a change to the baseline (GLCM)

JoostJM:
I split the commit (put the removal of SumVariance in a separate commit), as this constitutes a change to the baseline and therefore not just a documentation change.

PR #236, opened 2017-04-05
Add docker

blezek:
Added two extra Dockerfiles and associated documentation.


fedorov:
Another suggestion is to organize dockerfiles into a dedicated directory

Dockerfile.notebook
```diff
@@ -0,0 +1,40 @@
...
+# Install sample data and notebooks
+ADD data/ /home/jovyan/work/example_data/
+ADD bin/Notebooks/RadiomicsExample.ipynb /home/jovyan/work/
+ADD bin/Notebooks/FeatureVisualization.ipynb /home/jovyan/work/
+ADD bin/Notebooks/FilteringEffects.ipynb /home/jovyan/work/
+ADD bin/Params.yaml /home/jovyan/work/
+
+# Make a global directory and link it to the work directory
+RUN mkdir /data
+RUN ln -s /data /home/jovyan/work/data
```
jcfr: Instead of creating a custom image for this, I would suggest to:
* use the official jupyter images
* ask the user to simply clone the repository container the notebooks
* run the jupyter image mounting that folder
* Installation of Pyradiomics and its dependency would be step 1 of the notebook

Doing so, we would have one less image to maintain.

jcfr: The instruction would then be:
```
docker pull jupyter/datascience-notebook
git clone repo/containing/the-notebooks-dir
cd the-notebooks-dir
docker run -it --rm -p 8888:8888 -v $(pwd):/home/jovyan/work jupyter/datascience-notebook
```

README.md
```diff
@@ -73,6 +73,24 @@ To install this package on unix like systems run the following commands from the
 Detailed installation instructions, as well as instructions for installing PyRadiomics on Windows are available in the 
 [documentation](http://pyradiomics.readthedocs.io/en/latest/installation.html).
 
+### Docker
+
+PyRadiomics contains three "flavors" of [Dockers](https://www.docker.com/).  The first is a command line installation of the `pyradiomics` and `pyradiomicsbatch` command lines.  The second is a [Jupyter notebook](http://jupyter.org/) with PyRadiomics pre-installed with example Notebooks.  The third is a Ubuntu 16 based Docker for command line and interactive use. To build the Dockers:
+
+    docker build -t radiomics/cli .
+    docker build -t radiomics/notebook -f Dockerfile.notebook .
+    docker build -t radiomics/cli2 -f Dockerfile.ubuntu .
```
jcfr: The image should probably be named differently, it is not clear what is the difference between `cli` and `cli2`

Dockerfile.ubuntu
```diff
@@ -0,0 +1,19 @@
+# Build Pyradiomics inside Ubuntu 16.04
+
+FROM ubuntu:16.04
+
+MAINTAINER "Daniel Blezek" daniel.blezek@gmail.com
+
+# Install Scipy https://www.scipy.org/install.html
+USER root
+RUN apt-get update && apt-get install -y python-pip python-numpy python-scipy python-nose
```
jcfr: Some of these requirement are redundant. Using `pip install -r requirements.txt` should be sufficient to pull most of the requirements.

Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+  python get-pip.py && \
+  pip install wheel>=0.29.0 && \
+  pip install setuptools>=28.0.0 && \
+  pip install --trusted-host www.itk.org -f https://itk.org/SimpleITKDoxygen/html/PyDownloadPage.html SimpleITK>=0.9.1 && \
+  python -c "import SimpleITK; print('SimpleITK Version:' + SimpleITK.Version_VersionString())" && \
+  pip install -r requirements.txt && \
+  python setup.py install
+
+WORKDIR /usr/src
+ENTRYPOINT ["pyradiomics"]
```
jcfr: If the goal is to simply have an image providing pyradiomics, executable there is no need to include the compiler in the image to distribute. Instead, I would suggest to build a pyradiomics wheel built using manylinux. These would simply be installed in this image and it would be much smaller.

blezek:
@fedorov, @jcfr, Based on the comments, I'm going to propose that we settle down on 1 `Dockerfile`.  It will be based on the `jupyter/datascience-notebook` and install the example notebooks, as well as the `pyradiomics` command.  In the documentation, we can give examples on how to start up the notebook interface, and use the command in the Docker image.  I do realize the `datascience-notebook` is heavyweight, but it does come with a huge amount of helpful packages already baked in.  Having our examples (and more would be better) already baked in is a huge benefit.

We could also try to produce the smallest possible image for the command line version.

I hadn't planned on squashing the commits, it didn't seem to be the style in this repo.  Is it necessary?

fedorov:
Personally, I really really would hate the perfect to be the enemy of the good here. We can always improve later, but I think it is better to provide this functionality to the user early than have this PR pending much longer.

blezek:
@fedorov Happy to make changes, but I also agree with you.  We can iterate on Docker support.  I'll let someone else approve the PR.

fedorov:
@blezek would be great if you do those changes that you can and agree with today or tomorrow, we really would like to add docker support before the end of the week in time for the QIN meeting.

pieper:
+1 to making a full featured version available sooner and offering a
stripped down version later.

jcfr:
> offering a stripped down version later.

You are right size of the images is now critical problem.


That said, the other issues with the current approach is that we do not know if the created images will work or not. 

To keep up with the high bar maintained so far :smile: , I would hight recommend to add test(s) checking  that the built docker images works as expected. For example:
*  build and publish the image as part of the CI process (e.g execute a basic notebook that would import pyradiomics and do a simple computation or even run the complete test suite). The new CircleCi 2.0 makes working with docker even easier.

If not we should at least update the release check list to ensure the images are manually tested before each release.

Finally, for any given docker images ... we should be able to tracker what is the corresponding pyradiomics SHA version. This can be done easily adding metadata. See https://microbadger.com/labels#!
For example: https://github.com/QIICR/dcmqi/blob/f621a32eed0eadac852cee9860addbffb5d92460/docker/Makefile#L65-L72

I know it is tempting to move forward .. but I really think having systematic testing and provenance tracking for the docker images will pay off.

pieper:
Absolutely - testing is critical.  Let's make sure the user of the image
can verify that the image works as we intended.  A self test command that
the user can easily run would be very helpful.  Let's then add running the
tests to a notebook tutorial.

jcfr:
> A self test command that the user can easily run would be very helpful. Let's then add running the tests to a notebook tutorial.

Great. To achieve this, we could:

* bundle the tests as part of the package. Good news is that it is already the case. See https://github.com/Radiomics/pyradiomics/blob/913473e13fdd9657f5ce9e1c836cd24fac61e748/MANIFEST.in#L14

* install and execute nose.

* bonus: systematically executing all notebook should also be done.

With that, tests can be executed by anyone installing the package. And in that particular case, since docker provides a consistent and repeatable environment. We can really ensure the tests run in the environment we provide right after we build the docker images.

pieper:
Sounds awesome!

jcfr:
Great.

@blezek Let us know if you have any questions. I would be happy to review further.

blezek:
Hey @jcfr @pieper @fedorov & @JoostJM,

Bunch of commits, but I slimmed down to on Dockerfile that runs the example notebooks as part of the build with Python 2 and 3.  Also added a CircleCI v2 configuration that installs `radiomics` under Python 2 & 3, tests and executes the notebooks.

A working example build is at https://circleci.com/gh/blezek/pyradiomics/9.

If we configure DockerHub to auto build, we should have a nice pipeline going.

pieper:
Hi @blezek I checked out the branch and ran 'docker build' but got the error below.  

 ---> ebf7a1a7a6a3
Removing intermediate container 66d0c72f5b2d
Step 13 : ADD bin/Notebooks/FeatureVisualizationWithClustering.ipynb /home/jovyan/work/
lstat bin/Notebooks/FeatureVisualizationWithClustering.ipynb: no such file or directory


The file is referenced here, but looks like it's not in your branch.

https://github.com/Radiomics/pyradiomics/pull/236/files#diff-3254677a7917c6c01f55212f86c57fbfR33

blezek:
Sorry @pieper, I missed it.  Try again.  Need someone to look at that file and decide if it should be added to the Docker or not.

fedorov:
I confirm I could build docker without problems. Thank you Dan!

fedorov:
I am merging this so we officially have docker support in time for the QIN meeting that starts tomorrow.

fedorov:
@JoostJM you will need to configure docker hub yourself, it does not work for me, perhaps because of some permission issue. You will need to create Automated build, not Repository, which you have at the moment. I cannot do this, because somehow radiomics organization is not showing up in the list when I initiate creation of a new Automated build.

image

fedorov:
I was not able to access radiomics/pyradiomics organization from Docker hub because of the third-party access restriction policy. After disabling it, I can see the organization and the repo. However, I am still not able to create automated build under radiomics organization on Docker hub.

JoostJM:
@fedorov, I just added it. Setting it up now.

JoostJM:
PyRadiomics docker is building.

bin/Notebooks/PyRadiomics Example.ipynb
```diff
@@ -1,481 +0,0 @@
-{
```
JoostJM: @blezek, you deleted this example, was there a specific reason for this? This example was also used in the short video we made on PyRadiomics usage, and was therefore also made available in the repository

Issue #244, opened 2017-04-21
cmatrices lib path


pieper:
I did a fresh install of [the current version](https://github.com/Radiomics/pyradiomics/commit/0dd46a7bb5f401018f0e6b094afe7926dab026b2) on a linux VM (Amazon Linux AMI release 2016.09) following the installation instructions.

Then importing to python gives this error:

```
In [1]: import radiomics
Error loading C extensions, switching to python calculation:
Traceback (most recent call last):
  File "radiomics/__init__.py", line 233, in <module>
    from radiomics import _cmatrices as cMatrices
ImportError: cannot import name _cmatrices
```

As a workaround, this allows me to import:

`sys.path.append('/home/pieper/pyradiomics/build/lib.linux-x86_64-2.7/radiomics')
`

Should we add this to the documentation or is there a better way to solve this?

JoostJM:
@pieper, what is `radiomics.__path__`? It is possible/likely that this points to your repository (i.e. it's running in development mode). If so, first run `python setup.py develop` to make sure the C extensions are also available in source code.

I thought I included this in the FAQ, but apparently not, I'll correct this in a future PR.

pieper:
> what is radiomics.__path__?

```
In [2]: radiomics.__path__
Out[2]: ['radiomics']
```

Thanks @JoostJM , running `sudo python setup.py develop` resolves the issue.

I created a pull request to update the docs.  Can you confirm the wording is appropriate and merge?

JoostJM:
Resolved by #245 and #233

PR #245, opened 2017-04-21
Add extra install step

pieper:
Set up developer mode so cmatrices shared library is accessible.

Addresses [issue #244](https://github.com/Radiomics/pyradiomics/issues/244)

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -25,6 +25,9 @@ Installation on your system
   * ``cd pyradiomics``
   * ``sudo python -m pip install -r requirements.txt``
   * ``sudo python setup.py install``
+  
+  To use your build for interactive use and development:
+  * ``sudo python setup.py develop``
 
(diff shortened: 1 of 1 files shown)


docs/installation.rst
```diff
@@ -25,6 +25,9 @@ Installation on your system
   * ``cd pyradiomics``
   * ``sudo python -m pip install -r requirements.txt``
   * ``sudo python setup.py install``
+  
+  To use your build for interactive use and development:
+  * ``sudo python setup.py develop``
```
jcfr: Looks as it is consistent with previous step.


@JoostJM : I would then suggest to remove `sudo`, this is not a practice to encourage. Regular development should happen as regular user within a virtual environment.

JoostJM: @pieper, I agree with @jcfr. Can you remove the `sudo`'s?

JoostJM: @pieper, I'll merge this, and then remove the `sudo`'s in the update documentation PR (#233).

PR #252, opened 2017-05-03
Catch error in case of image/mask geometry mismatch

### LINKED ISSUE #251: checkMask will always fail if mask and image geometry do not match
checkMask essentially expects that image and mask geometry will match by calling Execute here: https://github.com/Radiomics/pyradiomics/blob/master/radiomics/imageoperations.py#L201

Thanks to Dev Gude for identifying and reporting this issue!

JoostJM:
Catch the error in checkMask and print out user-friendly error message. Then return None to prevent further feature extraction with this invalid image/mask combination.

Fixes #251


radiomics/imageoperations.py
```diff
@@ -198,7 +198,15 @@ def checkMask(imageNode, maskNode, **kwargs):
   logger.debug('Calculating bounding box')
   # Determine bounds
   lsif = sitk.LabelStatisticsImageFilter()
-  lsif.Execute(imageNode, maskNode)
+  try:
+    lsif.Execute(imageNode, maskNode)
+  except RuntimeError as e:
+    if "Both images for LabelStatisticsImageFilter don't match type or dimension!" in e.message:
+      logger.error('Image/Mask datatype or size mismatch. Potential solution: enable resampling')
```
fedorov: I think it would be very helpful to point to the specific section of documentation on how this can be done!

fedorov:
Would be good to also update the documentation in the same PR to clarify the instructions about how to implement the suggestions made in the error messages.

fedorov:
@JoostJM sorry, I am still missing the location in the documentation that specifies: 1) how to toggle resampling of the mask to the image raster; 2) how to change tolerance for SimpleITK geometry comparison operations. The comments contributed in the commit below provide explanation that would require modifying the code, or running an extra script. I thought you said it is possible to control those options from the yml configuration file - is this correct?

Also, where in the documentation all the parameters that can be passed using the configuration file described? I think it would be nice to have those in one location. What do you think?

JoostJM:
@fedorov, it is currently only possible to control resampling in general (i.e. resampling of both image and mask) from the yml file. However, I will add to parameters to control the global tolerance used by SimpleITK and a parameter to specify wheter to try to resample the mask if that geometry does not match.
The latter will 'fail' if the ROI in the mask contains areas outside the physical space of the image.

JoostJM:
Moreover, I am restructuring the documentation a bit, by adding documentation on customizing the extraction to the "usage" section, with additionally some documentation on setting up logging and how the parameter file works. Most of it is already documented, but at the start of the feature extractor module, which maybe not the best location.

JoostJM:
@fedorov, this last commit only represents the restructuring of the documentation, still working on the 2 paramters I want to add.

PR #253, opened 2017-05-08
STYLE: Correction of the Short Run Emphasis documentation

jbvimort:
Correction of a i that should be a j in the SRE formula in the documentation.

### DIFF
--- a/radiomics/glrlm.py
+++ b/radiomics/glrlm.py
@@ -248,7 +248,7 @@ class RadiomicsGLRLM(base.RadiomicsFeaturesBase):
 
     .. math::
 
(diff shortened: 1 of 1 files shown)


JoostJM:
@jbvimort  Good catch!

Btw don't worry about the failing circleci, that appears to be another bug. I'll take a look at that one.

JoostJM:
@Radiomics/developers 
CircleCi notebooks are currently failing because the Tumorbase testdata used there is downloaded from a URL that is currently down.
See also #254, where I changed the CircleCI tests to use the other notebooks, which make use of the test data included in the repository.

PR #254, opened 2017-05-09
Restructure folders

JoostJM:
Restructure directory tree to make a better distinction between additional scripts for special uses in PyRadiomics (`bin` folder) and examples (`examples` folder)

switch testing of notebooks to make use of the Hello* and PyRadiomicsExample Notebooks, which more extensively use the different parts of the toolbox and represent multiple ways in which the toolbox can be used.


JoostJM:
@Radiomics/developers Can this be merged? It also fixes a bug in the circleci testing, so I want to merge this PR first.

examples/Notebooks/FeatureVisualization.ipynb
```diff
@@ -377,7 +377,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython2",
-   "version": "2.7.12"
+   "version": "2.7.11"
```
jcfr: Not sure if this is needed.

jcfr:
Would it make sense to have:

`examples/batchExamples` -> `examples/examples`
`examples/Notebooks` -> `examples/notebooks`

and also:

`batchProcessingPandas.py` -> `batchProcessingWithPandas.py`
`batchprocessing.py` -> `batchProcessing.py`


Last, would it make sense to have all example in one folder ... 

```
examples/helloFeatureClass.py
[...]
examples/helloFeatureClassWithSettings.py  # This is where the exampleSettings would be used ... 
[...]
examples/batchProcessingWithPandas.py
examples/batchProcessing.py
...

```

fedorov:
Is there a good reason not to have notebooks as a top-level folder?

fedorov:
Overall, I don't have any major concerns, feel free to merge!

JoostJM:
@fedorov no particular reason, just a question of preference...
@jcfr, would it be better if the example settings are used with helloRadiomics. They are meamt to be used with the feature extractor...

jcfr:
> would it be better if the example settings are used with helloRadiomics. They are meamt to be used with the feature extractor...

Probably, something like `helloRadiomicsWithSettings.py` then.

I also wonder if the example should be numbered in the order they should be reviewed by a new user ..

```
0_helloFeatureClass.py
1_helloRadiomics.py
2_helloResampling.py
```

JoostJM:
In that case I would suggest to move helloFeatureClass should be moved to the back as it represents a non-standard use of pyradiomics

examples/Notebooks/FeatureVisualization.ipynb
JoostJM: No, but that is changed automatically if you run the notebooks.

JoostJM:
I restructured the folder, so that now notebooks is a separate top-level folder, and have one folder for all examples.

Seeing as the input CSV as it is structured for the batch examples is also the way it is structured for the batch script for the commandline, I moved the scripts showing how to build such a CSV from a dataset to the `bin` folder.

Moreover, I provided a test csv file to allow users to test the batch scripts using the test data included in the repository. This file is included with the examples.

Finally updated the notebooks some to better reflect recent changes.

JoostJM:
> I also wonder if the example should be numbered in the order they should be reviewed by a new user

I did not number the examples, as only the `helloRadiomics` example would be an example that a new user would have to view first. All the other examples show a specific use of PyRadiomics, but are more or less equal to each other in terms of when a new user would use them.

PR #261, opened 2017-05-30
Fix bug when creating symmetrical GLCM

JoostJM:
To make GLCM symmetrical, the python code performed an inplace addition: `P_glcm += numpy.transpose(P_glcm, (1, 0, 2))`. However, when the matrix exceeds a certain size, this produces buggy behaviour, where for some elements, the calculated value equals `P_glcm += P_glcm + numpy.transpose(P_glcm, (1, 0, 2))`. This is caused by the fact that numpy.transpose returns a *view*, and not a copy of the array. Then, if the additions have to be processed in multiple chunks, the addition will become erroneous.

Use `numpy.copy` to get a transposed copy of the GLCM and not just a view of the GLCM.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,14 @@ Release Notes
 Next Release
 ------------
 
+Bug fixes
+#########
+
+- In GLCM, the matrix is made symmetrical by adding the transposed matrix. However, ``numpy.transpose`` returns a view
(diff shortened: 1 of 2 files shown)


JoostJM:
Note: baseline is not affected, as this bug occurred when creating a symmetrical GLCM, whereas the baseline is made using the assymetrical GLCM.

PR #264, opened 2017-06-16
Replace SimpleITK's eigenvalue decomposition with a python one

JoostJM:
The eigenvalue decomposition calculation does not account for non-consecutive voxels in a line (along x dimension), causing the calculated values for the principal moments to be erroneous when a lesion has 'holes' in it. Furthermore, it assumes that all voxels on line x = i have identical y and z coordinates, but this is not necessarily the case.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,6 +19,10 @@ Bug fixes
   and not a copy of the array, causing erroneous results when adding it to the original array. use
   ``numpy.ndarray.copy`` to prevent this bug. **N.B. This affects the feature values calculated by GLCM when symmetrical
   matrix is enabled (as is the default setting).**
+- Use a python implementation to compute eigenvalues for ``shape.py`` instead of SimpleITK. The implementation in
+  SimpleITK assumes segmented voxels to be consecutive on the x-axis lines. Furthermore, it also assumes that all voxels
+  on a given line of x have the same values for y and z (which is not necessarily the case).
+  (`#264 <https://github.com/Radiomics/pyradiomics/pull/264>`_)
 
 -----------------
 PyRadiomics 1.2.0
--- a/radiomics/shape.py
+++ b/radiomics/shape.py
@@ -22,12 +22,6 @@ class RadiomicsShape(base.RadiomicsFeaturesBase):
 
     self.pixelSpacing = numpy.array(inputImage.GetSpacing()[::-1])
 
-    # Use SimpleITK for some shape features
-    self.logger.debug('Extracting simple ITK shape features')
-
-    self.lssif = sitk.LabelShapeStatisticsImageFilter()
-    self.lssif.Execute(inputMask)
-
     # Pad inputMask to prevent index-out-of-range errors
     self.logger.debug('Padding the mask with 0s')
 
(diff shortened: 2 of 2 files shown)


fedorov:
This looks like a serious problem with the ITK implementation.

cc: @thewtex @blowekamp

blowekamp:
Do you have a minimal example of the problem?

blowekamp:
For additional reference here is the Insight Journal article which describes how the principle moments calculation is done: http://hdl.handle.net/1926/584 There was also a follow up article which describes how the surface/perimeter computation is done: http://hdl.handle.net/10380/3342

I have made heavy use out of the principle axis computed with label maps. I recently extended the use of the principle axis with a computed [oriented bounding boxes](https://github.com/InsightSoftwareConsortium/ITK/commit/d0b6e92ef280a53d116cb4100175270ab0bdff60). And this is in the process of being added to SimpleITK SimpleITK/SimpleITK#189.

In my use cases and testing this computation has behavior correctly.  My experience has been that the computation takes into account the physical location of the pixels, and "gaps" in the labels. I would be very interested in a simple test case where the results are not correct.

JoostJM:
@blowekamp As soon as I can find the time, I'll prepare some examples and a detailed description for you. My statements are currently based on how the computation is done in ITK's [shapelabelobject](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/LabelMap/include/itkShapeLabelMapFilter.hxx). And of course the examples that led me to investigate this issue (some of the eigenvalues were negative).

I'll also take a look at the documents you sent and I will get back to you as soon as I can (currently at a congres though)

JoostJM:
@blowekamp, as a first thought, do your testcase included angled aquisitions? I believe this may be a larger influence and can even be the case if the algortihm takes gaps into account

blowekamp:
Here is a trivial test case, which seems to operate properly:
```
In [1]: import math

In [2]: import SimpleITK as sitk

In [3]: img = sitk.Image(5,5, sitk.sitkUInt8)

In [4]: img[1,2] = img[1,1] = 1

In [5]: stats = sitk.LabelShapeStatisticsImageFilter()

In [6]: stats.Execute(img); print stats.GetPrincipalAxes(1)
(1.0, 0.0, 0.0, 1.0)

In [7]: r = 0.78539816339

In [8]: img.SetDirection([math.cos(r), -math.sin(r), math.sin(r), math.cos(r)])

In [9]: stats.Execute(img); print stats.GetPrincipalAxes(1)
(0.7071067811918147, 0.7071067811812801, -0.70710678118128, 0.7071067811918146)
```

JoostJM:
@blowekamp, sorry for the late reply. 

I created a segmentation (for the brain1 testcase in PyRadiomics) with discontious segmentation, which seems to output meaningful values. However, when I change the direction to the direction I extracted from one of the cases where it failed. It also failed for the generated segmentation (i.e. produces negative eigenvalues).
Moreover, even when I use the original segmentation of breast1 testcase, it also fails when I change the direction.
I will prepare a representative example (main difference will be that this concerns a 3D volume, as opposed of a 2D one).

JoostJM:
@blowekamp, here are two examples. If rotated along one axis, values are still correct:

    import math
    import SimpleITK as sitk
    import numpy as np

    im_path = r'E:\Git-Repos\pyradiomics\data\breast1_image.nrrd'
    ma_path = r'E:\Git-Repos\pyradiomics\data\breast1_label.nrrd'
    image = sitk.ReadImage(im_path)
    mask = sitk.ReadImage(ma_path)

    r = math.radians(45.0)

    image.SetDirection([1.0, 0, 0, 0, math.cos(r), -math.sin(r), 0, math.sin(r), math.cos(r)])
    mask.SetDirection([1.0, 0, 0, 0, math.cos(r), -math.sin(r), 0, math.sin(r), math.cos(r)])

    lssif = sitk.LabelShapeStatisticsImageFilter()
    lssif.Execute(mask)
    print("Moments")
    print(lssif.GetPrincipalMoments(1))
    print("Axes")
    print(lssif.GetPrincipalAxes(1))

OUT:

Moments
(1.7488174696549337, 1.8303030713532475, 3.735485178452522)
Axes
(0.7950707161543119, -0.44533237368675166, 0.41175433605536305, -0.6065167008084678, -0.5840224148057925, 0.5394954222649374, 0.00021898465942798317, -0.6786728931900383, -0.7344406416415056)

However, if the volume is rotated along 2 axis:
breast1_rotated

Direction is set to 

    image.SetDirection([0.7950707161543119, -0.44533237368675166, 0.41175433605536305, -0.6065167008084678, -0.5840224148057925, 0.5394954222649374, 0.00021898465942798317, -0.6786728931900383, -0.7344406416415056])
    mask.SetDirection([0.7950707161543119, -0.44533237368675166, 0.41175433605536305, -0.6065167008084678, -0.5840224148057925, 0.5394954222649374, 0.00021898465942798317, -0.6786728931900383, -0.7344406416415056])

The output is:

Moments
(-254.61798391950742, 2.3911148490070246, 37.51723418381529)
Axes
(-0.6046568621057752, -0.7855033610930207, -0.1318125518301098, 0.37210010069808835, -0.1322641840296495, -0.9187206869791511, 0.7042241079151375, -0.6045582315497318, 0.3722603262457488)

The output from the numpy implementation in this PR is still correct (though not sorted yet. In PyRadiomics, results *are* sorted):

[ 3.73548518  1.74881747  1.83030307]
[[ 0.42892307  0.89225555  0.1410852 ]
 [ 0.56216099 -0.14139489 -0.81485122]
 [ 0.7071068  -0.42882108  0.5622388 ]]

blowekamp:
Thank you for demonstrating the problem. I was able to reproduce and have created a patch to address the issue here:
http://review.source.kitware.com/#/c/22472/1

JoostJM:
@blowekamp, going over the examples, if found out one last thing. It does not matter if it's rotated around just one or multiple axis, but whether it is rotated around the y and/or z axis (causing the x-part of the direction matrix to be different from "1, 0, 0" (as it is in the first example)). This is also logical if you look at the code and your fix; when not rotated or rotated around just the x-axis, the assumption that all voxels on the x-line have the same coordinates in y and z holds true.

PR #270, opened 2017-07-10
Remove circular import

JoostJM:
Currently,  PyRadiomics contains one circular import: `__init__.py` imports `base`, while `base` imports `radiomics` (`__init__.py`). However, base is only needed for the inheritance check in `getFeatureClasses`, change this to allow removal of the circular import.

While circular imports are allowed in python, then can cause unexpected behavior/bugs, and can mask an underlying bug preventing the import of a class (which is subsequently reported as a "cannont import name" error, instead of the actual bug preventing the import).

cc @Radiomics/developers

### DIFF
--- a/radiomics/__init__.py
+++ b/radiomics/__init__.py
@@ -12,7 +12,7 @@ import urllib
 
 import numpy  # noqa: F401
 
-from . import base, imageoperations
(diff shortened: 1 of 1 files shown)


pieper:
👍

PR #271, opened 2017-07-12
Update reference, move section customization

JoostJM:
Update the reference to reflect it is accepted for pulication in Cancer Research.

Move the section on customizing the extraction (and structuring the parameter file) to the top level to make it more visible.


pieper:
👍

Issue #268, opened 2017-07-05
Switch PyRadiomics License


JoostJM:
To have a more clear, but still open license, switch from the 3D slicer license to an open license.

No copyleft license, as this is incompatible with the Slicer license (as needed in SlicerRadiomics).

Alternatives:
- [Apache License 2.0](http://www.tldrlegal.com/license/apache-license-2.0-(apache-2.0))
- [MIT License](https://tldrlegal.com/license/mit-license)

cc @Radiomics/developers

JoostJM:
Also consider adding disclaimer specifying PyRadiomics is not intended for clinical use
 (e.g. print statement in the `__init__.py`)

fedorov:
Differences between the two alternatives, quoted from https://www.quora.com/Whats-the-different-between-Apache-v2-0-and-MIT-license:

![image](https://user-images.githubusercontent.com/313942/27871158-d7ca4934-6172-11e7-9aca-4cae480596a3.png)

JoostJM:
@haarburger, @jbvimort, @jaasantinha, @blezek, @vnarayan13.
I mentioned you here because you are also listed as contributors on PyRadiomics.

blezek:
+1 for MIT License.  Simple so people don't need to think about it.

fedorov:
Another option is [3-clause BSD](https://opensource.org/licenses/BSD-3-Clause), it is also used quite often in the research software. Very close to MIT, but the difference is this part:

> Neither the name of the copyright holder nor the names of its contributors may be used to endorse or promote products derived from this software without specific prior written permission.

fedorov:
Some relevant discussions 

* https://lists.opensource.org/pipermail/license-discuss/2008-July/015898.html
* license for XNAT: https://lists.opensource.org/pipermail/license-discuss/2011-April/016997.html (it looks like in the end XNAT people were convinced to go with the 2-clause BSD: https://www.xnat.org/download/)

Few more data points for the popular open source projects that came to my mind:
* [ITK is Apache-2](https://itk.org/gitweb?p=ITK.git;a=blob;f=LICENSE;h=d645695673349e3947e8e5ae42332d0ac3164cd7;hb=HEAD)
* [VTK is 3-clause BSD](https://github.com/Kitware/VTK/blob/master/Copyright.txt). 
* elastix [used to be simplified BSD, but then switched to Apache-2](http://elastix.isi.uu.nl/legal.php)
[...]

PR #272, opened 2017-07-13
Update license to 3-clause BSD

JoostJM:
Addresses #268
Change the license to 3-clause BSD

Question: should there be a explicit statement "Not intended for clinical use" and if so, where should it go?
Possibilities:
 - Extra clause in the license
 - Statement at the top of the documentation
 - Statement at the top of ReadMe.md
 - Disclaimer in the log. (as a warning, and as such, will also be part of the output to the command window in default mode).

cc @Radiomics/developers

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,199 +1,21 @@
+Copyright 2017 Harvard Medical School
 
-For more information, please see:
+Redistribution and use in source and binary forms, with or without modification, are permitted provided that the
+following conditions are met:
 
-                      http://www.slicer.org
+1. Redistributions of source code must retain the above copyright notice, this list of conditions and the following
+disclaimer.
 
-The 3D Slicer license below is a BSD style license, with extensions
-to cover contributions and other issues specific to 3D Slicer.
+2. Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following
+disclaimer in the documentation and/or other materials provided with the distribution.
 
+3. Neither the name of the copyright holder nor the names of its contributors may be used to endorse or promote products
+derived from this software without specific prior written permission.
 
-3D Slicer Contribution and Software License Agreement ("Agreement")
-Version 1.0 (December 20, 2005)
-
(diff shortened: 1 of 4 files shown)


fedorov:
> Question: should there be a explicit statement "Not intended for clinical use" and if so, where should it go?

@JoostJM no, I think it is not a good idea to add any extra clauses to the license. This will make the resulting license different from the 3-clause BSD, and will lead to a non-OSI-compliant license, obviating the purpose of license change. 

See pointers to the relevant discussions that I included in https://github.com/Radiomics/pyradiomics/issues/268#issuecomment-313195167. This specific issue you raise was discussed [here](https://lists.opensource.org/pipermail/license-discuss/2008-July/015899.html). Quoting the license text, "[...] FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED". "clinical use" is just one of the possible uses, all of which are already disclaimed.

pieper:
I agree with @fedorov we shouldn't change the wording of the license.  The "not intended for clinical use" should be in places like the log file, the readme, and a header or comment of the output files.

JoostJM:
@pieper, @fedorov, I added a disclaimer to the github readme and to the top of the documentation.

LICENSE.txt
```diff
@@ -1,199 +1,21 @@
+Copyright 2017 Harvard Medical School
```
fedorov: @JoostJM @hugoaerts @pieper thinking about this for dcmqi, I decided to list BWH as the copyright holder, see https://github.com/QIICR/dcmqi/blob/master/LICENSE.txt. The reason is the, in case of dcmqi, BWH is the recepient of the award from NIH. It might make sense to list DFCI here, since in the case of pyradiomics, DFCI is the recipient of the main award funding pyradiomics development.

hugoaerts:
I would keep it at HMS, as HMS shows it is more than only BWH or DFCI. Also, BWH is funding part of my lab, and some of these efforts helped the development of pyradiomics.

PR #273, opened 2017-07-14
Add example settings

JoostJM:
cc @Radiomics/developers 

Add example settings for different modalities as a starting point.
- [x] CT
- [x] MRI
- [ ] PET


JoostJM:
Merging this PR, example settings for PET can be added at a later point

PR #274, opened 2017-07-14
Revise style in base initialization

JoostJM:
Groups together common functions and splits assigning passed variables (such as `inputImage` and `inputMask`) and feature class percalculation (such as calculating the texture matrix), which enables customized initialization, such as in  `shape.py`.

Additionally, renamed matrixCoordinates to ROICoordinates, and moved assignment of the progress reporter to `__init__.py` (therefore only needing 1 call to a global function from the `base.py` module)

### DIFF
--- a/radiomics/base.py
+++ b/radiomics/base.py
@@ -61,12 +61,12 @@ class RadiomicsFeaturesBase(object):
       self.logger.warning('Missing input image or mask')
       return
 
-  def _initLesionWiseCalculation(self):
+  def _initSegmentBasedCalculation(self):
     self.imageArray = sitk.GetArrayFromImage(self.inputImage)
     self.maskArray = (sitk.GetArrayFromImage(self.inputMask) == self.label)  # boolean array
 
-    self.ROICoordinates = numpy.where(self.maskArray != 0)
-    self.size = numpy.max(self.ROICoordinates, 1) - numpy.min(self.ROICoordinates, 1) + 1
+    self.labelledVoxelCoordinates = numpy.where(self.maskArray != 0)
+    self.boundingBoxSize = numpy.max(self.labelledVoxelCoordinates, 1) - numpy.min(self.labelledVoxelCoordinates, 1) + 1
 
   def _applyBinning(self):
       self.matrix, self.binEdges = imageoperations.binImage(self.binWidth,
--- a/radiomics/base.py
+++ b/radiomics/base.py
@@ -6,7 +6,7 @@ import numpy
 import SimpleITK as sitk
 import six
 
-import radiomics
+from radiomics import getProgressReporter, imageoperations
 
 class RadiomicsFeaturesBase(object):
   """
@@ -41,15 +41,13 @@ class RadiomicsFeaturesBase(object):
   def __init__(self, inputImage, inputMask, **kwargs):
     self.logger = logging.getLogger(self.__module__)
     self.logger.debug('Initializing feature class')
-    # check if the handler to stderr is set and its level is INFO or lower
-    if logging.NOTSET < radiomics.handler.level <= logging.INFO:
-      self.progressReporter = radiomics._getProgressReporter
-    else:
-      self.progressReporter = radiomics._DummyProgressReporter
+    self.progressReporter = getProgressReporter
 
     self.kwargs = kwargs
     self.binWidth = kwargs.get('binWidth', 25)
     self.label = kwargs.get('label', 1)
+
+    self.coefficients = {}
 
     # all features are disabled by default
     self.disableAllFeatures()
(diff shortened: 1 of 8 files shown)


radiomics/base.py
```diff
@@ -63,13 +61,19 @@ def __init__(self, inputImage, inputMask, **kwargs):
...
       return
 
+  def _initLesionWiseCalculation(self):
     self.imageArray = sitk.GetArrayFromImage(self.inputImage)
-    self.maskArray = (sitk.GetArrayFromImage(self.inputMask) == self.label).astype('int')
+    self.maskArray = (sitk.GetArrayFromImage(self.inputMask) == self.label)  # boolean array
 
-    self.matrix = self.imageArray.astype('float')
-    self.matrixCoordinates = numpy.where(self.maskArray != 0)
+    self.ROICoordinates = numpy.where(self.maskArray != 0)
```
fedorov: Mybe `nonzeroVoxelCoordinates` or `labeledVoxelCoordinates` is a more intuitive name? When I think about ROI, I think either a box, or a segmented region, and it is not immediately clear what `ROICoordinates` means.

radiomics/base.py
```diff
@@ -63,13 +61,19 @@ def __init__(self, inputImage, inputMask, **kwargs):
...
+    self.size = numpy.max(self.ROICoordinates, 1) - numpy.min(self.ROICoordinates, 1) + 1
```
fedorov: maybe `boundingBoxSize` or `ROIBoundingBoxSize` is a better name

radiomics/base.py
```diff
@@ -63,13 +61,19 @@ def __init__(self, inputImage, inputMask, **kwargs):
       self.logger.warning('Missing input image or mask')
       return
 
+  def _initLesionWiseCalculation(self):
```
fedorov: Not immediately clear from the name of the function what is its intent. When I just look at the name, I think that it means it can take an image with multiple lesions, and calculate features for each one, but it does not look like that's what it is actually doing. May also make sense to add some documentation about the function, especially considering it is reimplemented in derived classes.

radiomics/firstorder.py
```diff
@@ -30,6 +30,12 @@ def __init__(self, inputImage, inputMask, **kwargs):
     self.pixelSpacing = inputImage.GetSpacing()
     self.voxelArrayShift = kwargs.get('voxelArrayShift', 0)
 
+    self._initLesionWiseCalculation()
+
+  def _initLesionWiseCalculation(self):
+    super(RadiomicsFirstOrder, self)._initLesionWiseCalculation()
+    self.targetVoxelArray = self.imageArray[self.ROICoordinates].astype('float')
+
     self.logger.debug('Feature class initialized')
```
fedorov: For consistency, may be more helpful to have logging message `<feature class name> feature class initialized` for each of the feature classes?

radiomics/base.py
JoostJM: will change to `labeledVoxelCoordinates`

radiomics/base.py
JoostJM: will change to `boundingBoxSize`

radiomics/firstorder.py
JoostJM: I'll add it, but in the log the logger is already feature class specific, (i.e. logger would be `radiomics.firstorder`). The reason I'll still add it is because the logger name is not necessarily printed (as for example in the standard output to the console when verbosity is high enough.

radiomics/base.py
JoostJM: This function prepares the feature class to calculate features (i.e. this function calculates the texture matrix, upon which the features are calculated).
The reason for the name is so that there can be a second, parallel function for voxel-wise calculation, which can be called `_initVoxelWiseCalculation`.

I'll add a docstring, but because this is a private function, this will only be visible in the source code, not the documentation generated by sphinx.

fedorov: Oh, I now see what you meant. Then perhaps you meant to call it `_initLesionBasedCalculation` as opposed to `_initPixelBasedCalculation`. I think we need a consult from a native speaker - calling for help @pieper!

pieper: I agree with @fedorov's point about it lesion-wise sounds like per-lesion, so lesion-based would be a better fit.

But why are we using the term 'lesion' at all here?  It doesn't show up much at all in the master except in the faq and maybe we want to use a more generic term like 'segment' or 'structure'?  As we discussed with Hugo today on the call radiomics might be used to analyze organs or other structures (e.g. hippocampus).   Perhaps we can create a style guide that defines the terms 'mask' 'boundingBox' and 'segment' so we use them consistently.  Here mask and boundingBox are both aspects of the 'lesion' or 'segment' level concept (i.e. I don't think we say it explicitly but I believe we always mean: pyradiomics calculates features of segments, where segments often correspond to segmentations of lesions.)

fedorov: > But why are we using the term 'lesion' at all here? 

Great point. 

This may also be an opportune time to start promoting the use of term "segment" in the radiomics community. The caveat to this is that we already have different concepts that are described by the same name "segment". Slicer "segment" and DICOM segmentation image "segment" are not the same, and both are mostly unknown in the broader community.

pieper: True, there's no perfect wording that would harmonize with all other systems, but we can define some reasonable terminology and use it consistently in this codebase.

fedorov: Yes, I agree.

PR #277, opened 2017-07-21
BUG: outlier removal not working

fabianbalsiger:
Hi guys,

I think the PR is self-explanatory. The if had no influence at all.

Cheers,
Fabian

Cc: @Radiomics/developers

### DIFF
--- a/radiomics/imageoperations.py
+++ b/radiomics/imageoperations.py
@@ -554,6 +554,7 @@ def normalizeImage(image, scale=1, outliers=None):
 
     newImage = sitk.GetImageFromArray(imageArr)
     newImage.CopyInformation(image)
+    image = newImage
 
   image *= scale
 


pieper:
LGTM

Thanks for the fix!

tatsunidas:
Hi, @JoostJM 

How to test "removeOutliers not working" correctly?

I tried following,

```
# The dataset was IBSI PAT1 CT (ibsi_1_ct_radiomics_phantom).
# 
settings = {}
settings['binCount'] = 32
settings['binWidth'] = None
# If enabled, resample image (resampled image is automatically cropped.
settings['resampledPixelSpacing'] = None  # [3,3,3] is an example for defining resampling (voxels with size 3x3x3mm)
settings['interpolator'] = sitk.sitkBSpline
settings['label'] = 1 # 
settings['removeOutliers'] = float(0.5) # and did testing using with 1, 2, 3. 

firstOrderFeatures = firstorder.RadiomicsFirstOrder(image, mask, **settings)
firstOrderFeatures.enableFeatureByName('Mean', True)

results = firstOrderFeatures.execute()
print('Calculated first order features: ')
for (key, val) in six.iteritems(results):
  print('  ', key, ':', val)
```

Here, if I change removeOutliers to 0.5, 1, 2, 3..., a mean value was not changing(in this case, result was always -46. This value did not apply removeOutliers).

I think, maybe my step is something wrong.
I appreciate it if you let me know a suitable procedure or thanks for your suggestion to my misunderstanding. 

tatsuaki

PR #280, opened 2017-08-01
BUG: initialize inputImage from kwargs

fedorov:
'inputImage' parameter passed to the constructor in kwargs was ignored

### DIFF
--- a/radiomics/featureextractor.py
+++ b/radiomics/featureextractor.py
@@ -54,6 +54,8 @@ class RadiomicsFeaturesExtractor:
       if len(kwargs) > 0:
         self.logger.info('Applying custom settings')
         self.settings.update(kwargs)
+        if 'inputImage' in kwargs.keys():
+          del self.settings['inputImage']
       else:
         self.logger.info('No customized settings, applying defaults')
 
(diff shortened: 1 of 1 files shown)


pieper:
Do we have an example or a test where we send in inputImage as a kwarg?  What type is it expected to be?

Here self.inputImages is a dictionary with a field Original that is a dictionary.  Would that also be expected to come as the inputImage arg?  Everywhere else in the code we use inputImage to refer to an sitkImage.

fedorov:
If I use the parameter file, this works:

```
inputImage:
  Original: {}
  LoG: {'sigma': [5]}
```

It is initialized from kwarg here: https://github.com/Radiomics/pyradiomics/blob/master/radiomics/featureextractor.py#L211

fedorov:
To me, it is confusing, but at this point I am not trying to improve the nomenclature, just get things to work!

JoostJM:
>It is initialized from kwarg here: >https://github.com/Radiomics/pyradiomics/blob/master/radiomics/featureextractor.py#L211

@fedorov, the fact that the parameter file already worked is handled [elsewhere](https://github.com/Radiomics/pyradiomics/blob/master/radiomics/featureextractor.py#L137-L140)

The reason input images was ignored when present in `kwargs` was due to the fact that `kwargs` corresponds to the settings, which is only 1 out of 3 customization categories present in the parameter file (`setting`, `inputImage` and `featureClass`). When featureclass is initiatied with `kwargs`, features and input images had to be set *after* initialization, for example by a call to [`enableInputImageByName()`](https://github.com/Radiomics/pyradiomics/blob/master/radiomics/featureextractor.py#L175).

Let's discuss this on Wednesday.

Issue #286, opened 2017-08-15
CircleCI Tests failing


JoostJM:
Currently, CircleCI tests are failing, as it cannot find the python2 environment anymore.

I think this is related to this commit [here](https://github.com/jupyter/docker-stacks/commit/32b3d2bec23bc46fab1ed324f04a0ad7a7c73747), where python 2 support is phased out of jupyter/datascience-notebook

JoostJM:
related issue in jupyter/docker-stacks: [#432](https://github.com/jupyter/docker-stacks/issues/432)

JoostJM:
Fixed by #287

PR #287, opened 2017-08-16
Fix missing python 2 environment

JoostJM:
See also [#433](https://github.com/jupyter/docker-stacks/pull/433) on jupyter/docker-stack.
Python 2 support is now removed from jupyter/datascience-notebook. Update dockerfile and config.yml to install python2 kernel manually.


JoostJM:
Fixes #286
cc @Radiomics/developers

PR #291, opened 2017-08-21
Revert 280 and 281, update customization names and documentation

JoostJM:
Reverts #280 and #281, as these PRs introduced a change which confuses the distinction between the 3 types of customization in PyRadiomics.

Furthermore, renames variables and updates documentation to make this distinction more clear.
Throughout PyRadiomics, "parameters" refers to one or more categories of customization in PyRadiomics (i.e. all customization possible). The 3 categories of parameters are:
1. imageType: specifies what image types (original and/or derived using filters) to extract features from.
2. featureClass: specifies which feature classes (and optionally which individual features) to extract from the enabled image types.
3. setting: specifies the settings used to customize how the features are extracted.

In the parameter file, all 3 categories can be specified. When initializing the feature extractor without a parameter file, or when using the feature classes directly, only type 3 customization ("setting") can be provided. Changes of type 1 and 2 are applied *after* intialization, through dedicated function calls (e.g. `enableAllFeatures` and `enableAllImageTypes` in feature extractor).

Includes updates to the documentation to further clarify this, and include a small example for the structure of the parameter file.

cc @Radiomics/developers


notebooks/PyRadiomicsExample.ipynb
```diff
@@ -367,33 +369,33 @@
...
-      "\t original_glcm_Energy : 0.00290880217681\n",
-      "\t original_glcm_SumSquares : 39.9781084143\n",
-      "\t original_glcm_ClusterProminence : 26251.1709801\n",
-      "\t original_glcm_Entropy : 8.79428086119\n",
-      "\t original_glcm_Imc2 : 0.692033706271\n",
-      "\t original_glcm_Imc1 : -0.091940840043\n",
-      "\t original_glcm_DifferenceAverage : 5.58932678922\n",
-      "\t original_glcm_Id : 0.276140402104\n",
-      "\t original_glcm_ClusterTendency : 103.142793792\n",
+      "\t original_glcm_SumVariance : 108.731393255\n",
```
fedorov: It is not obvious from the title and comments in the PR that it should lead to a change in the feature baselines. Is this expected?

fedorov:
@JoostJM note that SlicerRadiomics uses https://github.com/Radiomics/pyradiomics/tree/slicer branch, you will need to make the corresponding change there too.

fedorov:
Other than the two comments above, looks good to me, great to see improved documentation!

notebooks/PyRadiomicsExample.ipynb
JoostJM: @fedorov, only the values for GLCM are different in the notebook, other values have just been moved.
As to the difference in GLCM values, that has to do with the bug that was fixed in #261, which could cause the GLCM to be incorrect, depending on your hardware and size of the input.

It took me a while to find out, as I got correct values (equal the 'new' values in this PR) on my regular computer, even when using PyRadiomics version `1.1.1.post80.dev0+g215030b`. Turns out that this computer is big enough to add the transposed matrix in one go for brain1 (therefore not causing bug as in #261), but my laptop isn't.

Baseline for GLCM was generated using my regular computer, therefore these were correct, even prior to #261, but the notebooks were generated using my laptop, with different values for GLCM as a result (last update of this notebook was prior to #261).

JoostJM:
> @JoostJM note that SlicerRadiomics uses https://github.com/Radiomics/pyradiomics/tree/slicer branch, you will need to make the corresponding change there too.

True, I will merge this branch into the slicer branch (I didn't want to make 2 PRs, I'll reference this PR in the merge commit)

PR #297, opened 2017-08-23
ENH: add a dockerfile that exposes command line tools

fedorov:


JoostJM:
@fedorov I could not link to your docker file in a separate repo under radiomics organisation (docker repo name must be the same as github repo name), but I was able to add an additional build trigger (tag CLI). However, the build failed. Can you investigate?

fedorov:
I am looking into it. It seems as if the support for the version of CentOS in the base package is no longer available, or something like this, but updating the version I ran into another issue. I will make another PR when resolved.

PR #300, opened 2017-09-04
Remove Sum Variance

JoostJM:
Remove Sum Variance (GLCM), as this feature is mathematically equal to Cluster Tendency. Document the proof of this equality in a new documentation section (Excluded Radiomic Features).

As this proof is coded in LaTeX, here is a render of the proof:
proof_sumvar_clusten

Aside from this equality, there is also a near equality of Joint Average and Sum Average  (when GLCM is symmetrical, Sum Average = 2 * Joint Average). Therefore, disable Sum Average by default in the example parameter files and add a warning to the docstring of Sum Average.

Finally, there was a small error in the documentation of the GLCM formulas (u_x and u_y erroneously represented as a function of i and j, respectively (they are scalar values)).

cc @Radiomics/developers


clarehchao:
Hi all,

Thanks for making this nice repo for extracting radiomic features with such
ease and flexibility!

I was going through the documentation on how the GLCM-related features are
defined and found some inconsistency that I was wondering if someone can
help clarify since this pull request was regarding the feature definition.

Sum variance:

   - sum variance was defined slightly differently in Aert et al., 2014
   (Nature Communications, supplementary documentation) as the following


[image: Screen Shot 2017-09-04 at 2.22.08 PM.png]
Instead of SA (sum average that was used in pyradiomics), sum entropy (SE)
was used instead. Also, sum variance was defined using sum entropy as well
in this this link
<http://earlglynn.github.io/RNotes/package/EBImage/Haralick-Textural-Features.html>.
Although it makes more sense to use sum average since we're calculating sum
variance, perhaps the source of definition that I mentioned may be
incorrect. so please clarify if you can!


   - Difference variance
      - in pyradiomics documenttaion, difference variance was defined as
      the following

[image: Screen Shot 2017-09-04 at 3.08.11 PM.png]
but in radiomics/glcm.py, line 485-488, difference variance is defined as
diffvar = numpy.sum((pxSuby * ((kValuesDiff[:, None] - diffavg) ** 2)), 0)

which translate to this equation instead..
\textit{difference variance} = \displaystyle\sum^{N_g-1}_{k=0}{(k
-DA)^2p_{x-y}(k)}

Let me know what you think and I am happy to help out by submitting a pull
request if needed. Thanks a lot for your time and help!

Cheers,
Clare Huang

JoostJM:
@clarehchao, thanks for pointing out the error in the difference average formula! I already fixed it with a commit in this PR, as I also corrected some other errors in GLCM formula documentation, so another PR is not necessary.

As to your remark on Sum Variance, we are aware of this difference. Indeed, PyRadiomics used to have 2 variants of Sum Variance, but this has been removed as of PR #233, (specifically commit 2042e27). The implementation as defined in the paper by Aerts et al., as well as on the site you mention use the definition as it is in the original Haralick paper (Haralick, R., Shanmugan, K., Dinstein, I; Textural features for image classification; IEEE Transactions on Systems, Man and Cybernetics; 1973(3), p610-621). However, the use of Sum Entropy is most likely the result of a typo (`f8` instead of `f6` in the Sum Variance formula `f7`), as a variance is a sum of the difference between intensity and mean squared.

This choice for Sum Variance using the Sum Average is also consistent with the definition by the Imaging Biomarker Standardisation Initiative (IBSI). Their feature definitions are available [here](https://arxiv.org/abs/1612.07003).

clarehchao:
Thank you for the clarification!!! Keep pressing on making awesome package
like pyradiomics!

Clare

Issue #303, opened 2017-09-27
Memory Error during Feature Extraction (for many images or large mask)


michaelschwier:
When I tried to extract features on many images in a loop using PyRadiomics I ran into a memory error. I was able to reproduce two types of memory errors with a simple script and just one image and mask, repeatedly calling the feature extraction on the same image.

I put the code to reproduce the error [here](https://gist.github.com/michaelschwier/893125cbec9bcb4c6ce43bbf9fa411d9) on gist
You can find the test image data to run the script (as well as the script file itself again) [here](https://www.dropbox.com/sh/dq51zg7orizqbfk/AAA12m0n2VyiGU03haYknYlha?dl=0)

There are two different error cases:
1. Trying to extract features from a rather large mask (-> "MemoryError" in numpy)
2. Trying to extract features from small masks for many images in a loop (-> "Failed to allocate memory for image" in SimpleITK)

You can reproduce them by (un)commenting on of the lines defining which mask to use (see also the top explanatory comment in the script).

Case 1 error details:
```
Traceback (most recent call last):
  File "d:/Test/PyRadiomicsMemoryExceptionTest.py", line 26, in <module>
    featureVector = extractor.execute(testImage, testMask, label = 1)
  File "C:\Python36\lib\site-packages\pyradiomics-1.2.0.post25.dev0+g13274ff-py3.6-win32.egg\radiomics\featureextractor.py", line 354, in execute
    shapeClass = self.featureClasses['shape'](croppedImage, croppedMask, **self.settings)
  File "C:\Python36\lib\site-packages\pyradiomics-1.2.0.post25.dev0+g13274ff-py3.6-win32.egg\radiomics\shape.py", line 62, in __init__
    physicalCoordinates -= numpy.mean(physicalCoordinates, axis=0)  # Centered at 0
  File "C:\Python36\lib\site-packages\numpy\core\fromnumeric.py", line 2909, in mean
    out=out, **kwargs)
  File "C:\Python36\lib\site-packages\numpy\core\_methods.py", line 54, in _mean
    arr = asanyarray(a)
  File "C:\Python36\lib\site-packages\numpy\core\numeric.py", line 583, in asanyarray
    return array(a, dtype, copy=False, order=order, subok=True)
MemoryError
```

Case 2 error details:
```
Traceback (most recent call last):
  File "d:/Test/PyRadiomicsMemoryExceptionTest.py", line 27, in <module>
[...]

PR #310, opened 2017-10-09
Remove 32 bit support

JoostJM:
When using PyRadiomics in 32 bits architecture, memory errors can arise when extracting large masks, or many small ones (see also #303).
Therefore, remove support for 32 bits architecture by raising an exception when trying to install PyRadiomics on 32 bits architecture.

Update documentation accordingly.
Reformat `setup.py` script to adhere to editorconfig settings

cc @Radiomics/developers, @michaelschwier

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ Furthermore, an instruction video is available [here](http://radiomics.io/pyradi
 
 ### Installation
 
-PyRadiomics is OS independent and compatible with both Python 2.7 and Python >=3.4.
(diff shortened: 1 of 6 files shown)


jcfr:
👍 

What do you think of logging an Exception if a 32-bit runtime is detected ?

jcfr:
👍 

What do you think of logging an Exception if a 32-bit runtime is detected ?

fedorov:
@jcfr there is one already: https://github.com/Radiomics/pyradiomics/pull/310/files#diff-2eeaed663bd0d25b7e608891384b7298R16

Issue #283, opened 2017-08-03
Installation failure on Windows


fedorov:
As reported by Wei Mu, installation is failing on Windows as shown below:

![image](https://user-images.githubusercontent.com/313942/28938370-e0bea9b8-785b-11e7-9b0e-dba037354c4e.png)

Documentation does not list any special requirements for Windows: http://pyradiomics.readthedocs.io/en/latest/installation.html.

JoostJM:
The error listed here appears to be an issue with the compiler set-up (so not a problem of the pyradiomics code). The traceback appears to be going through a set of python scripts before ending up in one that looks to me to be an interface to cygwin compiler, but then hits a null reference exception on a version check, so maybe because cygwin is not installed, or that python is unable to find it.

I myself compile using microsoft VC, which also has free distributions for python compiling. See [here](https://wiki.python.org/moin/WindowsCompilers).

fedorov:
Note that there are no instructions about compiler set-up in the pyradiomics documentation: http://pyradiomics.readthedocs.io/en/latest/installation.html

![image](https://user-images.githubusercontent.com/313942/29274035-6eab9ea4-80d4-11e7-8d7c-70fe0997bd24.png)

JoostJM:
As of PyRadiomics 2.0, Wheels are automatically built and uploaded for PyRadiomics releases. This removes the need for local compiling.

PR #311, opened 2017-10-11
setup: use scikit-build for building cmatrices and cshape C extensions

jcfr:
See #283

This commit introduces new dependencies: "scikit-built" and "cmake".

scikit-build is a drop-in replacement to setuptools.setup function
allowing to easily compile and package extensions (C/C++/Cython) by
bridging CMake and setuptools. See http://scikit-build.org

CMake is is an open-source, cross-platform family of tools designed
to build, test and package software. See https://cmake.org

Currently, scikit-build and cmake have to be explicitly (see [1]) installed
on the system. This could be done by simply doing:

  pip install -U scikit-build cmake

How does it work ?
------------------

In addition to simplifying setup.py, two new file have been added:

```
<root>
  |
  |---- CMakeLists.txt                          (1)
  .
  .
  .
  |--- ...src
  |        |--- CMakeLists.txt                  (2)
  .
  .
```

The first CMakeLists.txt specifies requirements for

* Python interpreter,

* the associated python libraries

* and a "PythonExtension" modules containing convenience CMake
  functions allowing to easily build extension by encapsulating
  system introspection and platform specific logic.

and include the subdirectory associated with the other CMakeLists.txt.

The second CMakeLists.txt is specific to "cmatrices" and "cshape"
extensions and it specifies:

* requirement for NumPy

Then, it declares:

* libraries for each extension, associate python extension specific properties

* and finally add an install rule specifying where in the package
  hierarchy the extension should be installed.

Et voila,

[1] Note that improvement to python packaging will be available shortly and
it will be possible to include scikit-build and cmake directly in pyproject.toml
See here for more details: https://www.python.org/dev/peps/pep-0518/

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,9 @@
+cmake_minimum_required(VERSION 3.7)
+
+project(radiomics)
+
+find_package(PythonInterp REQUIRED)
+find_package(PythonLibs REQUIRED)
+find_package(PythonExtensions REQUIRED)
+
+add_subdirectory(radiomics/src)
--- a/radiomics/src/CMakeLists.txt
+++ b/radiomics/src/CMakeLists.txt
@@ -1,0 +1,11 @@
+find_package(NumPy REQUIRED)
+
+add_library(_cmatrices MODULE _cmatrices.c cmatrices.c)
+python_extension_module(_cmatrices)
+target_include_directories(_cmatrices PRIVATE ${NumPy_INCLUDE_DIR})
+
+add_library(_cshape MODULE _cshape.c cshape.c)
+python_extension_module(_cshape)
+target_include_directories(_cshape PRIVATE ${NumPy_INCLUDE_DIR})
+
+install(TARGETS _cmatrices _cshape LIBRARY DESTINATION radiomics)
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,10 +35,14 @@ jobs:
           name: Install pyradiomics in Python 2 and 3
           command: |
             source activate python2
+            python -m pip install -U cmake
+            python -m pip install -U scikit-build
             python -m pip install --no-cache-dir -r requirements.txt
             python -m pip install --no-cache-dir -r requirements-dev.txt
             python setup.py install
             source activate root
+            python -m pip install -U cmake
+            python -m pip install -U scikit-build
             python -m pip install --no-cache-dir -r requirements.txt
             python -m pip install --no-cache-dir -r requirements-dev.txt
             python setup.py install
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,7 @@ docs/_build/
 
 # PyBuilder
 target/
+
+# scikit-build
+_skbuild
+MANIFEST
--- a/scikit-ci.yml
+++ b/scikit-ci.yml
@@ -28,9 +28,8 @@ install:
   commands:
     - python --version
     - python -m pip install --disable-pip-version-check --upgrade pip
-    - $<RUN_ENV> pip install wheel>=0.29.0
-    - $<RUN_ENV> pip install setuptools>=38.6.0
-    - $<RUN_ENV> pip install numpy>=1.9.2
+    - $<RUN_ENV> pip install -U scikit-build
+    - $<RUN_ENV> pip install -U cmake
     - $<RUN_ENV> pip install --trusted-host www.itk.org -f https://itk.org/SimpleITKDoxygen/html/PyDownloadPage.html SimpleITK>=0.9.1
     - $<RUN_ENV> python -c "import SimpleITK; print('SimpleITK Version:' + SimpleITK.Version_VersionString())"
     - $<RUN_ENV> pip install -r requirements.txt
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,6 @@ import platform
 import sys
 
 from setuptools.command.test import test as TestCommand
-
 import versioneer
 
 try:
--- a/setup.py
+++ b/setup.py
@@ -1,13 +1,22 @@
 #!/usr/bin/env python
 
-from distutils import sysconfig
+from __future__ import print_function
+
 import platform
 import sys
 
-import numpy
-from setuptools import Extension, setup
 from setuptools.command.test import test as TestCommand
+
 import versioneer
+
+try:
+    from skbuild import setup
+except ImportError:
+    print('scikit-build is required to build from source.', file=sys.stderr)
+    print('Please run:', file=sys.stderr)
+    print('', file=sys.stderr)
+    print('  python -m pip install scikit-build')
+    sys.exit(1)
 
 # Check if current PyRadiomics is compatible with current python installation (> 2.6, 64 bits)
 if sys.version_info < (2, 6, 0):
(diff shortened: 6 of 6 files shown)


fedorov:
@jcfr I tried `pip install cmake` just to try on a linux system, but I am getting this error:

image

I have pip installed with pyenv.

jcfr:
@fedorov  Not sure what's happening. Look like the `PATH` is not properly updated because it reports issue with an older version of CMake.

May you could try to run:

```
$ hash -r
$ pyenv rehash
```

Otherwise, here is what I tried to reproduce the problem ... 

```
$ git clone https://github.com/pyenv/pyenv.git ~/.pyenv
Cloning into '/home/jcfr/.pyenv'...
remote: Counting objects: 15540, done.
remote: Compressing objects: 100% (39/39), done.
remote: Total 15540 (delta 33), reused 61 (delta 27), pack-reused 15468
Receiving objects: 100% (15540/15540), 2.76 MiB | 0 bytes/s, done.
Resolving deltas: 100% (10625/10625), done.
Checking connectivity... done.
$ export PYENV_ROOT="$HOME/.pyenv"
$ export PATH="$PYENV_ROOT/bin:$PATH"
$ eval "$(pyenv init -)"
$ pyenv install 3.5.0
Downloading Python-3.5.0.tar.xz...
-> https://www.python.org/ftp/python/3.5.0/Python-3.5.0.tar.xz
Installing Python-3.5.0...
WARNING: The Python bz2 extension was not compiled. Missing the bzip2 lib?
WARNING: The Python readline extension was not compiled. Missing the GNU readline lib?
Installed Python-3.5.0 to /home/jcfr/.pyenv/versions/3.5.0

$ pyenv shell 3.5.0

$ pyenv versions
  system
* 3.5.0 (set by PYENV_VERSION environment variable)

$ pip install --upgrade pip
Collecting pip
  Using cached pip-9.0.1-py2.py3-none-any.whl
Installing collected packages: pip
  Found existing installation: pip 7.1.2
    Uninstalling pip-7.1.2:
      Successfully uninstalled pip-7.1.2
Successfully installed pip-9.0.1
$ pip install cmake 
Collecting cmake
  Downloading cmake-0.8.0-cp35-cp35m-manylinux1_x86_64.whl (19.2MB)
    100% |████████████████████████████████| 19.3MB 101kB/s 
Installing collected packages: cmake
Successfully installed cmake-0.8.0
$ cmake --version 
cmake version 3.9.0

CMake suite maintained and supported by Kitware (kitware.com/cmake).

```

I can also confirm shims are used:

```
$ which python 
/home/jcfr/.pyenv/shims/python
```

jcfr:
@JoostJM  In order to nicely support packaging and distributing pyradiomics, I just rebased this topic.

jcfr:
@fedorov I will also ensure it can be installed and packaged in the SlicerRadiomics extension

JoostJM:
@jcfr, It appears to be working now on Windows and Mac, but not on Linux (appears to be missing a file, I'll try to follow-up on this). However, even though it works on the Appveyor, I'm still getting the unresolved externals on my own windows machine. Could this be due to the fact that I use Visual Studio 2015 compiler? Do you have an idea why this doesn't work with skbuild, but does work with native python setuptools?

jcfr:
> Could this be due to the fact that I use Visual Studio 2015 compiler? 

This should not be a problem 

> Do you have an idea why this doesn't work with skbuild, but does work with native python setuptools?

Again, what are the step you follow and command line you type ?

JoostJM:
This is my setup output

[setup_cmake.txt](https://github.com/Radiomics/pyradiomics/files/1724571/setup_cmake.txt)

jcfr:
In which environment and which command to you run to get this output ? 

Anaconda, windows terminal, Visual studio terminal ... 

And which command (e.g python setup.py bdist_wheel, python setup.py install, ...)

JoostJM:
regular windows terminal with compiler environment variables set

JoostJM:
@jcfr, I debugged the Linux build. By default on my home Linux machine, it sort of performs a "develop" build, where the extensions are built and placed in the source directory, and in the python site-packages it just updates the `easy-install.pth` file. This works and PyRadiomics works without issue.

However, on CircleCI, it installs it in the site-packages, and then I discovered that skbuild ignores the `package_data` keyword argument in the `setup()` call. This argument specifies some additional files required by PyRadiomics (the parameter script validation files), which are therefore not copied and missing at runtime. This is the source of the CircleCI error in this branch.

JoostJM:
@jcfr, I think I found the error on Windows. On one machine it didn't work, but it did on the other and the issue (as it is so often) is the architecture of the builds. Apparently on the machine where it didn't work, the generator defaulted to Visual Studio 2015, but not Visual Studio 2015 Amd64. And because my python build is 64 bits, it failed. 

Is there some way to check if a generator is for 64- or 32-bits compiling? And can that be cross-checked with the target architecture?

JoostJM:
@jcfr, I tried to reinstall using the current master of scikit-build, and it works now for python 2.7 on my Windows machine. However, for python3.5, there is an issue where CMake appears to find some non-existend libs and subsequently fails the build because it cannot link to those libraries. I attached the output from my commandline. The 'libraries' in question are "optimized" and "debug" (sound more like a designator, as there both an optimized and debug library are also found.
It is possible that this is not Python 3 specific, but has more to do with the fact that for python 3 I also installed the debug libraries, but did not do so for python 2.

If I go to the generated CMake files and update the project (by removing those 2 references), the project builds without error.

[setup output](https://github.com/Radiomics/pyradiomics/files/1806759/setuperr.log)

jcfr:
Thanks for the report.

We will soon release a new version of scikit-build that should allow to successfully build pyradiomics. If I recall, this commit https://github.com/scikit-build/scikit-build/commit/c3a19bacd93898b929faee8c99966aa1503567de ensure datafiles are packaged as expected.

jcfr:
> The 'libraries' in question are "optimized" and "debug" (sound more like a designator, as there both an optimized and debug library are also found.

This should be addressed by https://github.com/scikit-build/scikit-build/pull/306/files

JoostJM:
> Thanks for the report.
>
> We will soon release a new version of scikit-build that should allow to successfully build pyradiomics. If I recall, this commit scikit-build/scikit-build@c3a19ba ensure datafiles are packaged as expected.

Yes, that's why I first built scikit-build from source (current master), which already incorporates this change.

CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+cmake_minimum_required(VERSION 3.7)
+
+project(radiomics)
+
+find_package(PythonInterp REQUIRED)
+find_package(PythonLibs REQUIRED)
+find_package(PythonExtensions REQUIRED)
+
+add_subdirectory(radiomics/src)
```
jcfr, line 9: After transitioning to `scikit-build-core`

```
cmake_minimum_required(VERSION 3.15...3.26)

project(${SKBUILD_PROJECT_NAME} VERSION ${SKBUILD_PROJECT_VERSION} LANGUAGES C)

find_package(Python COMPONENTS Interpreter Development.Module NumPy REQUIRED)

add_subdirectory(radiomics/src)
```

radiomics/src/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
...
+
+add_library(_cmatrices MODULE _cmatrices.c cmatrices.c)
+python_extension_module(_cmatrices)
+target_include_directories(_cmatrices PRIVATE ${NumPy_INCLUDE_DIR})
+
+add_library(_cshape MODULE _cshape.c cshape.c)
+python_extension_module(_cshape)
+target_include_directories(_cshape PRIVATE ${NumPy_INCLUDE_DIR})
+
+install(TARGETS _cmatrices _cshape LIBRARY DESTINATION radiomics)
```
jcfr, line 11: After transitioning to `scikit-build-core`

```
Python_add_library(_cmatrices MODULE _cmatrices.c cmatrices.c)
target_link_libraries(_cmatrices PRIVATE Python::NumPy)

Python_add_library(_cshape MODULE _cshape.c cshape.c)
target_link_libraries(_cshape PRIVATE Python::NumPy)

install(TARGETS _cmatrices _cshape DESTINATION radiomics)
```

hjmjohnson:
The newest build uses "scikit-build-core>=0.11.2".  These changes are no longer relavant.

PR #321, opened 2017-10-24
Remove duplicate features

JoostJM:
In addition to the equality between the formulas for Sum Variance and Cluster Tendency (#300), the following features also have identical formulas:

GLDM - Gray Level Non-Uniformity Normalized = First Order - Uniformity
GLCM - Dissimilarity = GLCM - Difference average
GLCM - Homogeneity1 = GLCM - ID (Exactly identical formula, only difference in name)
GLCM - Homogeneity2 = GLCM - IDM (Exactly identical formula, only difference in name)

Remove these features from the feature classes and update the documentation accordingly.
Remove features from example parameter files where applicable.
Add mathematical proof for these equalities to the documentation.

Simplify formulas in documentation for GLSZM, GLRLM and GLDM (use a predifined variable Nz for the sum of the matrix).


fedorov:
I wonder what IBSI people think about this. On one hand, indeed, not good to have duplicate features, but on the other hand, someone can read a paper using (for example) dissimilarity feature, or read the IBSI document, and it is not obvious to map it back. An alternative idea could be to have a setting that disables duplicate features. What do you think?

pieper:
If there are publications that use different terminology for the same formula then I'd think we would choose a canonical name for the formula and then include at least documentation about the mapping (and maybe a stub function with a deprecation warning).  From what I can tell this PR removes all reference to the removed features which could be a problem for current users.

jcfr:
> then include at least documentation about the mapping 

👍

JoostJM:
> From what I can tell this PR removes all reference to the removed features which could be a problem for current users.

Not entirely, All removed features are documented in their [own section](http://pyradiomics.readthedocs.io/en/latest/removedfeatures.html), which also includes the mathematical proof of why they are equal to other features (with links to those features). Moreover, I added notes with those other features, specifying that those features are equal, with pointers to this extra section in the documentation.

JoostJM:
What I can do is add a deprecation warning and raise a specific exception there (which will then not include the results in the featurevector) how about a [NotImplementedError](https://docs.python.org/2/library/exceptions.html#exceptions.NotImplementedError)?

fedorov:
> All removed features are documented

Right, but I think the sad reality is that most users don't read documentation (even when things don't work!). So I would definitely vote for a stub function. A problem with stub function, specifically for radiomics studies, is that I would expect users might often run extraction in batch mode and ignore all warnings, but perhaps it is a reasonable middle ground. They will at least have a chance to be notified.

fedorov:
> the only thing these functions do is warn the user that they are deprecated (by raising and catching a DeprecationWarning exception)

+ printing the URL of the page summarizing the duplicated features and the rationale for excluding them?

JoostJM:
> printing the URL of the page summarizing the duplicated features and the rationale for excluding them?

Yes, it will display/log a warning containing the feature name, the equal other feature name and a reference to http://pyradiomics.readthedocs.io/en/latest/removedfeatures.html.

JoostJM:
@fedorov, @pieper, I now added the duplicate features as stub functions, returning a deprecation warning. Furthermore, I also added a `deprecated` decorator to mark those functions (and some other functions that are related to other features (e.g. firstorder standard deviation; related to firstorder variance). This will prevent them from being enabled by default.

For those feature functions that are marked, but do not return a deprecation warning (features that are similar, but not equal to other features), can still be calculated when they are explicitly enabled (i.e. specified by name in the parameter file).

fedorov:
👍

pieper:
👍

PR #338, opened 2018-01-12
Add PyRadiomics Models

JoostJM:
The first commit only defines an example for the model configuration file and adds some code to validate these using PyKwalify:
- Update the PyKwalify validation in PyRadiomics to also allow validation of PyRadiomics model files.
- Add an example model file to explain the structure.
- Update requirements.txt to enforce PyKwalify version != 1.6.0, as this version contains a bug that breaks the usage of partial schemas, which is utilized to allow validation of the two use cases (parameter file and model file) **N.B. in the previous version (1.5.2) PyKwalify contains a bug that resets the output to stdout upon loading the package. This means that when using jupyter notebooks, the output does not show in the notebook, but in the commandline running jupyter!**
- Because partial schemas are used, load the schemas outside of pykwalify and pass them during validation as a dict. (Enables customized schema dict reflecting the two use cases)
- Update the testParams script to allow testing of both use cases.

TODO:
- [ ] add functionality to apply a model (i.e. extract features, parse results and call model specific modules to apply the actual model.
- [ ] add several modules containing model specific code (for example to apply linear or logistic regression)
- [ ] design unit tests for the new functionality
- [ ] update test-settings test to allow testing of model configuration files
- [ ] add command line interface to make new functionality available to the user
- [ ] update documentation
- [ ] build slicer module to make new functionality available in slicer (not in this PR)

cc @Radiomics/developers

### DIFF
--- a/bin/testParams.py
+++ b/bin/testParams.py
@@ -3,25 +3,55 @@
 # custom parameters specified will be printed. If validation fails, an error message specifying cause of validation
 # error will be printed.
 
-import sys
+import argparse
 
 import pykwalify.core
 
 from radiomics import getParameterValidationFiles
 
-def main(paramsFile):
-  schemaFile, schemaFuncs = getParameterValidationFiles()
 
-  c = pykwalify.core.Core(source_file=paramsFile, schema_files=[schemaFile], extensions=[schemaFuncs])
+def main(paramsFile, is_model=False):
+  if is_model:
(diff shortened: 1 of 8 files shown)


hjmjohnson:
@JoostJM  This is 7 years old and must be re-implemented on top of the master branch.   Closing this implementation.  Please re-open if effort will be made to implement these changes on the master branch.

PR #347, opened 2018-02-22
ENH: Refactor CLI scripts

JoostJM:
Combine the `pyradiomics` and `pyradiomicsbatch` entry point into 1 joint entry point `pyradiomics`.
This new entry point operates in batch mode when input to `Image|Batch` has `.csv` extension.
In batch mode, `Mask` argument is ignored. In single mode `Mask` inputs is required.

Additionally, enables easy multi-threaded extraction by specifying `--jobs` / `-j` argument (with integer indicating number of parallel threads to use.
Will only work in batch mode, as extraction is multi-threaded at the case level (1 thread per case).

Removes argument `--label` / `-l`, specifying an override label to use is now achieved through specifying `-s "label:N"`, with N being the label value.

Also includes a small update to initialization of the feature extractor:
- In addition to a string pointing to a parameter file, feature extractor now also accepts a dictionary (top level specifies customization types 'setting', 'imageType' and 'featureClass') as 1st positional argument
- Regardless of initialization with or without a customization file/dictionary, `kwargs` passed to the constructor are used to override settings.

cc @Radiomics/developers


pieper:
Maybe for now include a stub entrypoint for pyradiomicsbatch that issues an error and pointer to the documentation for switching to the new style.

pieper:
lgtm 👍

PR #348, opened 2018-02-23
ENH: Nosetest for wavelets on different ROIs

michaelschwier:
Follow up on PR #346.

This test check equality of features extracted after wavelet filtering on the same image but different ROIs (with padding large enough such that wavelet filter doesn't run into the image border for pixels under the mask).

Test data is a small and simple generated image containing a checkered sphere and noise around.


tests/test_wavelets.py
```diff
@@ -0,0 +1,45 @@
...
+    settings = {}
+    settings['binWidth'] = 10
+    extractors.append(self.getTestExtrator(settings))
+    settings['preCrop'] = True
+    settings['padDistance'] = 6
+    extractors.append(self.getTestExtrator(settings))
+
+    features = []
+    for extractor in extractors:
+      for data in self.testData:
```
JoostJM: These two loops sort of specify different parameters of testing. In PyRadiomics we split these out as well. 
This is acchieved by the `@parameterized.expand(<generatorFunction>())`, where `<generatorFunction>` is a function that utilizes the `yield` statement (i.e. is a python generator). Inside that function are the loops to define your testcases, and the `yield` statement returns an equal number of arguments as are consumed by your test function.
See [here](https://github.com/Radiomics/pyradiomics/blob/master/tests/test_features.py#L50) for an example.

michaelschwier: The loops don't specify different test cases/scenarios. It is just one case which compares 4 different ROI variations for computational equality. Parameterized creates independent test cases which wouldn't help. Only thing I could do is create parameterized test cases that check always two pairs of my 4 variations. I don't see the benefit though :/

michaelschwier: Also, we could think about parameterizing the features, so one case for each feature. However, the actual intention is to check the wavelet filter for equality, using the feature values is just a means to make it easier. To implement the intention more correctly I could re-write this to actually get the wavelet filtered images and compare the pixels under the mask for equality ... but I am not sure if it's worth the effort since this does the job.

tests/test_wavelets.py
```diff
@@ -0,0 +1,45 @@
...
+        features.append(extractor.execute(data["image"], data["mask"], label=1))
+
+    dfFeatures = pd.DataFrame(features)
+    for column in dfFeatures:
+      if not column.startswith("general"):
```
JoostJM: @michaelschwier, you can prevent the addition of the general info columns by calling `addProvenance(False)` on each extractor when you instantiate it in `getTestExtractor()`

JoostJM:
Superseded by #387

Issue #354, opened 2018-03-06
Errors and crash with specific dataset


fedorov:
With the attached dataset, there are several things I cannot explain:
* if I run it with the image and label, I get runtime errors in shape feature calculation (no config file, default parameters). However, if I run it passing label as both the image and label, it works fine. I expected shape feature calculation should not depend on the image content.
```
/Users/fedorov/.pyenv/versions/3.5.3/Python.framework/Versions/3.5/lib/python3.5/site-packages/pyradiomics-1.3.0.post60+g1af2eff-py3.5-macosx-10.12-x86_64.egg/radiomics/shape.py:446: RuntimeWarning: invalid value encountered in sqrt
  return numpy.sqrt(self.eigenValues[0] / self.eigenValues[2])
/Users/fedorov/.pyenv/versions/3.5.3/Python.framework/Versions/3.5/lib/python3.5/site-packages/pyradiomics-1.3.0.post60+g1af2eff-py3.5-macosx-10.12-x86_64.egg/radiomics/shape.py:416: RuntimeWarning: invalid value encountered in sqrt
  return numpy.sqrt(self.eigenValues[0]) * 4
```
* when the error happens, pyradiomics does not exit cleanly, and appears to consume a lot of resources for a long time before dying.

The dataset was provided by the user to reproduce the issue discussed in [this Slicer forum thread](https://discourse.slicer.org/t/radiomics-module-crashes-anytime-i-use-a-theta-parameter-map/2152).

[Archive.zip](https://github.com/Radiomics/pyradiomics/files/1786041/Archive.zip)

JoostJM:
@fedorov, I received the same error regardless both when I used image and mask and when I used mask alone. 

The cause is in a machine-precision error. The segmentation in the dataset represents a 2D segmentation, and therefore the smallest of the 3 eigenvectors should be 0. But because we transform to physical coordinates (taking into account rotation and scaling), this actually computed to ~-7.9e-16. 

This in turn caused the error in `numpy.sqrt`, but again, my windows PC simply calculates `nan` without crashing.

Still, I implemented a 3-fold fix.
1) Because we do not need the eigenVectors, we can also calculated the eigenvalues using just the scaled coordinates, without rotation. In case of a 2D segmentation, this then results in correct calculation of 0 for the smallest eigenvector.
1) After computation of eigen values. They are checked if they are negative, but > -1e-10. If that is the case, a warning is logged and they are rounded to 0.
[...]

PR #355, opened 2018-03-07
BUG: Fix machine precision errors in eigen value calculation

JoostJM:
If the segmentation in the dataset represents a 2D segmentation, the smallest of the 3 eigenvectors should be 0.
But because a transform to physical coordinates (taking into account rotation and scaling), this can actually compute very small (~1e-16) negative numbers.

This in turn causes an error in numpy.sqrt. This commit implements a 3-fold fix.

1) Because we do not need the eigen vectors, we can also calculate the eigenvalues using just the scaled coordinates, without rotation. In case of a 2D segmentation, this then results in correct calculation of 0 for the smallest eigen value.
2) After computation of eigen values. They are checked if they are negative, but > -1e-10. If that is the case, a warning is logged and they are rounded to 0.
3) Add additional checks to ensure the eigenvectors are not < 0 (effectively < -1e-10) in the feature functions. If that is the case, a warning is logged and nan is returned.


JoostJM:
fixes #354

PR #375, opened 2018-04-17
-swt3 comments

tommydino93:
I have just added some comments to the method _swt3 which is called inside the method getWaveletImage


JoostJM:
@tommydino93, It appears some of your changes break the testing (more specifically, `flake8`, which ensures a consistent coding style). Can you take a look at the error messages and correct as necessary?

tommydino93:
Alright @JoostJM, I think it's ok now!

PR #386, opened 2018-06-04
ENH: Add support for using a fixed bin count

JoostJM:
For extra compatibility allow user to extract using a fixed bincount. 
Include a warning when fixed bincount is used to show the recommendation for a fixed bin width.

See [this FAQ topic](http://pyradiomics.readthedocs.io/en/latest/faq.html#why-is-there-no-parameter-to-specify-a-fixed-bin-count) for more details.


fedorov:
I have to say, as a user, I could be confused reconciling the guidelines from IBSI (see below from v6 p.12) and strong recommendation not to use fixed bin number in pyradiomics.

image

In any case, good to see this feature, since we had questions about it several times now!

Issue #388, opened 2018-06-08
trimesh package in docker container for LBP3D filter


cpwardell:
First; thanks for writing this software, it's excellent and much appreciated.

If you attempt to use the LBP3D filter in the docker image downloaded using:
`docker pull radiomics/pyradiomics`

You get the following error:
`Could not load required package "scipy" or "trimesh", cannot implement filter LBP 3D`

I solved it by installing the trimesh package using pip. 
`pip install trimesh`

Please update your dockerfile/dockerhub.

JoostJM:
@cpwardell, thanks for the tip! I'll update the dockerfile to fix this issue.

jonasharad:
I would be grateful for quick feedback on whether the solution above: "I solved it by installing the trimesh package using pip.
pip install trimesh" also refers to pyradiomics integrated into 3D Slicer? Cheers, Marko

Please update your dockerfile/dockerhub.

PR #389, opened 2018-06-13
BUG: Fix missing packages in Dockerfile

JoostJM:
To enable local binary pattern 2D and 3D, scipy, scikit-image and trimesh are additionally required, but these are not mentioned in the requirements (to allow installing without these packages).

To enable these filters in the CLI-docker, add pip install lines for these packages to the dockerfile.

### DIFF
--- a/docker/cli/Dockerfile
+++ b/docker/cli/Dockerfile
@@ -93,6 +93,9 @@ RUN git clone https://github.com/radiomics/pyradiomics.git && \
   python get-pip.py && \
   pip install wheel>=0.29.0 && \
   pip install setuptools>=28.0.0 && \
+  pip install scipy && \
+  pip install trimesh && \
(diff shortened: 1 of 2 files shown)


JoostJM:
This update builds successfully on dockerhub (tag Test): [Docker container](https://hub.docker.com/r/joostjm/pyradiomics/)

JoostJM:
Fixes #388

fedorov:
@JoostJM it looks good to me, but why don't we just list those packages in `requirements.txt`? I am probably missing something basic here, right?

JoostJM:
@fedorov, the reason I don't want to list them in requirements.txt is because scipy is very difficult to install on windows. So I wrote the code in such a way that it works without those packages, except for the for the part using those packages of course.

PR #391, opened 2018-06-20
ENH: Add check to prevent too many bins from being generated

JoostJM:
When the GLCM is calculated in C, an array is generated of size Ng x Ng x Na (Ng = number of bins, Na = number of angles). If the number of bins exceeds 12852 with 13 angles, the resulting array has more than 2,147,483,648 elements, which in turn causes the index variable (int32) to overflow, crashing the python interpreter.

Seeing as generally ~100 bins are advised, limit the number of bins instead of using datatypes that can reach higher numbers. A maximum number of 7000 is chosen (allowing for a maximum of 43 angles in GLCM). If the number of bins exceeds this number, an error is raised.

Additionally, add a size check prior to instantiating the output arrays in C. If the size > 2147483648, throw an error.

Fix typo in error message when calculation of angles fails.

### DIFF
--- a/radiomics/imageoperations.py
+++ b/radiomics/imageoperations.py
@@ -47,6 +47,10 @@ def getBinEdges(binwidth, parameterValues):
   if len(binEdges) == 1:  # Flat region, ensure that there is 1 bin
     binEdges = [binEdges[0] - .5, binEdges[0] + .5]  # Simulates binEdges returned by numpy.histogram if bins = 1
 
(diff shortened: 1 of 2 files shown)


JoostJM:
Related to [this issue](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/pyradiomics/y2lEgOFKWOg) on the pyradiomics google groups.

Issue #49, opened 2016-04-26
Stabilize dynamic binning for Wavelet transform output


vnarayan13:
Recompute bin edges and binwidth to maintain consistent bin count for all normalized stationary wavelet transform outputs.

sha168:
I’m trying to extract voxel-wise features from the original MR image and its wavelet transforms, but I’m struggling with enabling the “dynamicBinning”.

In my .yaml-file I have tried the following:

imageType:
  Original: {}
  Wavelet: {}
featureClass: 
  glcm:
    - 'Autocorrelation'
    - 'JointAverage'
setting:
  normalize: true
  normalizeScale: 100
  interpolator: 'sitkBSpline'
  resampledPixelSpacing: [2, 2, 2]
  binWidth: 60
  dynamicBinning: true

  voxelArrayShift: 300
  label: 1
  force2D: false

voxelSetting:
  kernelRadius: 1
  maskedKernel: true
  initValue: nan
  voxelBatch: 10000

(am I supposed to provide the "binWidth" parameter for the original image, like I have done here?) However, this gives me the following error code:

 --- All found errors ---
["Key 'dynamicBinning' was not defined. Path: '/setting'"]
Traceback (most recent call last):
  File "<input>", line 152, in <module>
  File “/Users/…/radiomics/featureextractor.py”, line 59, in __init__
    self._applyParams(paramsFile=args[0])
  File “/Users/…/featureextractor.py”, line 158, in _applyParams
    params = c.validate()
  File “/Users/…/core.py”, line 167, in validate
    error_msg=u'.\n - '.join(self.validation_errors)))
pykwalify.errors.SchemaError: <SchemaError: error code 2: Schema validation failed:
 - Key 'dynamicBinning' was not defined. Path: '/setting'.: Path: '/'>

Did I make a mistake in the parameter-file? I’m using version 2.1.2.post61+g3efae04.

sha168:
Ok, thanks for the quick reply! Is the bug addressed in this issue [https://github.com/Radiomics/pyradiomics/issues/456](url) fixed in this branch?

JoostJM:
@sha168, I just rebased the branch on the current master, so if you've checked out the branch with commit af7f32b, then yes.

hjmjohnson:
Assuming this is no longer a relevant issue.

PR #414, opened 2018-08-24
ENH: Implement dynamic binning

JoostJM:
Add the possibility to enable dynamic binning, which scales the bin width by the ratio between the range of intensities seen in the derived and original images (only including intensities in the ROI). This is only done if dynamic binning is enabled (new parameter `dynamicBinning`, boolean, default false) and no custom bin width has been defined for that filter.

This also has no effect if a fixed bin count is used.

Addresses issue #49.

cc @Radiomics/developers


hjmjohnson:
@JoostJM This is 7 years old and must be re-implemented on top of the master branch.   Closing this implementation.  Please re-open if effort will be made to implement these changes on the master branch.

PR #420, opened 2018-08-28
Extend resegmenting functionality

JoostJM:
In addition to the original mode (now default mode "absolute"), add new modes "relative" and "sigma", which allow for a dynamic threshold defined as either a threshold relative to the maximum value found in the ROI ("relative") or as a n * sigma difference from the mean ("sigma").

Additionally, it is now allowed to provide either 1 or 2 thresholds, where in case of 1 threshold it is interpreted as the lower threshold (i.e. retaining everything equal to or higher than that threshold).

In case the function fails, a ValueError is raised. The checks to determine the validity of the resegmented mask is not determined in this function, but is applied in the pipeline defined by featureextractor.

### DIFF
--- a/radiomics/featureextractor.py
+++ b/radiomics/featureextractor.py
@@ -45,6 +45,8 @@ class RadiomicsFeaturesExtractor:
 
     self.featureClasses = getFeatureClasses()
 
+    self.generalInfo = None
+
     self.settings = {}
     self._enabledImagetypes = {}
     self._enabledFeatures = {}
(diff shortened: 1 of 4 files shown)


JoostJM:
This PR now also contains some updates for the provenance info and the behavior in case of failing checks:
The general info class is now updated at multiple points during the extraction, allowing for information reflecting the input data in various stages of pre-processing

Instead of logging an error and returning an empty feature vector when checks fail, raise a ValueError detailing the cause of the failing check.

PR #427, opened 2018-09-20
Mesh volume and diameters

JoostJM:
To increase consistency between surface area, volume and maximum diameters, calculate all 3 on the same triangle mesh generated using the marching cubes algorithm.

Original definition of volume (No of voxels * volume of 1 voxel) is retained as `ApproximateVolume`.

This also matches the definition of shape features as they are defined by IBSI.

cc @Radiomics/developers


fedorov:
Since both surface areas are approximations, how about renaming them to `MeshSurfaceArea` and `VoxelSurfaceArea`?

fedorov:
Or maybe even better - `TriangulatedSurfaceArea`.

radiomics/shape.py
```diff
@@ -239,57 +235,38 @@ def getMaximum3DDiameterFeatureValue(self):
...
-    .. warning::
-      This feature is only available when C Extensions are enabled
+    Maximum 2D diameter (Row) is defined as the largest pairwise Euclidean distance between tumor surface mesh
+    vertices in the column-slice (usually the sagittal) plane.
     """
-    if self.diameters is None:
-      self.diameters = self.calculateDiameters()
     return self.diameters[2]
 
   def getMajorAxisFeatureValue(self):
```
fedorov: Should this be `MajorAxisLength`? To me, "major axis" implies vector.

JoostJM: @fedorov, good point! But I'll change that in a separate PR, as it is unrelated to the changes proposed here.

JoostJM:
> Since both surface areas are approximations, how about renaming them to `MeshSurfaceArea` and `VoxelSurfaceArea`?

@fedorov, certainly a good suggestion (although I propose we do this for the 2 volume features, as there is only 1 surface area feature, the mesh-based one. i.e. `MeshVolume` and `VoxelVolume`. Or maybe `VoxelbasedVolume` / `VoxelwiseVolume`?

> Or maybe even better - `TriangulatedSurfaceArea`.

Certainly also a possibility, although I'm more partial to `Mesh`, `Triangulated` makes me think of triangulation. Plus, `MeshVolume` is easier to pronounce... @pieper, @hugoaerts, any preferences?

JoostJM:
@fedorov, do you also want me to change the names of `Maximum3DDiameter` to `Maximum3DMeshDiameter`, or maybe `MeshDiameter-3D` (with `3D` replaced by `2DColumn`, `2DRow` and `2DSlice` for the others?

pieper:
I like the idea of prefixing the measurement with Mesh or Voxel (e.g. MeshVolume, MeshSurfaceArea and VoxelVolume and VolumeSurfaceArea) for clarity and consistency.  Same logic would imply MeshMaximum3DDiameter, and other metrics.

fedorov:
This sounds fine to me. I suggested Triangulation in place of Mesh because to be precise - voxel surface is also a mesh. But I agree Triangulation is too long, and the distinction can be explained in the documentation.

JoostJM:
@fedorov, @pieper, changing all feature names to explicitly state they are Mesh or Volume really lengthens the names. Therefore, I have for now just updated the documentation. There it is now specified that all shape features, unless otherwise specified, are based on the mesh approximation.

Only exceptions are VoxelVolume (which I did explicitly rename, to contrast it with MeshVolume) and the PCA-derived features (major, minor and least axis lengths, elongation and flatness), but for those I explicitly mention they are not based on the mesh.

pieper:
@JoostJM yes, it makes sense to have the human readable names short and easy to use.

@fedorov when you express these in SR, are there modifiers to differentiate the 'from mesh' vs the 'from voxel' variants?  For consistency probably the 'from voxel' modifier should be used for the cases Joost mentioned even if there isn't a 'from mesh' variant just so consumers of the document can know without referring to the documentation.

fedorov:
I think it is ok to not have that modifier explicitly in the name. We can't expect deriving fully qualified definition of the feature calculation from just the name in the general case. We can handle mapping as needed for SR encoding via a separate lookup table or some decorators in the function documentation.

PR #442, opened 2018-10-19
Add 2D Shape features

JoostJM:
Add a new featureclass implementing shape features describing a 2D shape.
This class can only be extracted if `force2D = True` and the size of the bounding box is 1 in the dimension specified by `force2Ddimension` (i.e. it must be a truly 2D segmentation)

This featureclass is therefore not enabled by default.

cc @Radiomics/developers

### DIFF
--- a/data/baseline/baseline_shape2D.csv
+++ b/data/baseline/baseline_shape2D.csv
@@ -1,0 +1,42 @@
+featureName,brain1_shape2D,brain2_shape2D,breast1_shape2D,lung1_shape2D,lung2_shape2D
+diagnostics_Image-original_Maximum,3057.0,1523.0,248.0,3071.0,1671.0
+diagnostics_Mask-original_CenterOfMass,"(43.13548018292683, 20.012385670731703, 26.000000000000014)","(57.44906387665198, -22.31087860241864, -10.700000762939446)","(-99.99007990243902, -1.789887073170732, 28.35000000000003)","(-21.721181556195972, -120.73608609510089, -612.4999999999998)","(53.84355387019346, -97.05089679126613, -249.9999999999999)"
+diagnostics_Mask-corrected_VolumeNum,1,1,1,1,1
+diagnostics_Mask-corrected_Size,"(256, 256, 25)","(256, 256, 23)","(45, 120, 21)","(512, 512, 48)","(512, 512, 61)"
+diagnostics_Mask-corrected_Mean,859.3841463414634,422.5947136563877,131.1829268292683,-51.518731988472624,17.4510550996483
+diagnostics_Image-original_Minimum,0.0,0.0,6.0,-1024.0,-1024.0
+diagnostics_Mask-corrected_Spacing,"(0.7812499999999999, 0.7812499999999999, 6.499999999999998)","(0.7812499999999999, 0.7812499999999999, 6.499999999999998)","(0.664062, 0.664062, 2.1)","(0.5703125, 0.5703125, 5.0)","(0.6269531, 0.6269531, 5.0)"
+diagnostics_Versions_Numpy,1.15.1,1.15.1,1.15.1,1.15.1,1.15.1
+diagnostics_Mask-corrected_CenterOfMass,"(43.13548018292683, 20.012385670731703, 26.000000000000014)","(57.44906387665198, -22.31087860241864, -10.700000762939446)","(-99.99007990243902, -1.789887073170732, 28.35000000000003)","(-21.721181556195972, -120.73608609510089, -612.4999999999998)","(53.84355387019346, -97.05089679126613, -249.9999999999999)"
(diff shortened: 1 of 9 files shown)


fedorov:
> the size of the bounding box is 1 in the dimension specified by force2Ddimension (i.e. it must be a truly 2D segmentation)

I would say the input is "truly 2D" if the ITK image passed as input is a 2D image, or perhaps if the spacing in z dimension is 0. Otherwise the definition can be ambiguous.

pieper:
I don't think a spacing of zero is a good idea.  I would think we can come up with a reasonable definition of grid samples that gracefully handles any combination of dimensions and spacings in any dimension without becoming degenerate.

fedorov:
Definitely - I think it is better if the tool allowed processing of true 2d images. I suggested 0 z spacing in case we want to maintain the current convention that the input is always a 3d ITK image.

JoostJM:
> I would say the input is "truly 2D" if the ITK image passed as input is a 2D image, or perhaps if the spacing in z dimension is 0. Otherwise the definition can be ambiguous.

Very true, and would nicely help those users that use 2D input formats (e.g. PNG; though it does not address the multiple channels issue). On the other hand, it is likely that there are also users who have 3D input image data, but only single-slice segmentations. Those are also really 2D shapes (even though the real-world shape is 3D), but there it is more likely that dimensionality of the input data is 3D.

By the way, the latter also holds true for 2D DICOM (e.g. x-ray, mammography). ITK will read this in as 3D data, with size 1 in a certain dimension.

JoostJM:
> I would say the input is "truly 2D" if the ITK image passed as input is a 2D image, or perhaps if the spacing in z dimension is 0. Otherwise the definition can be ambiguous.

Very true, and would nicely help those users that use 2D input formats (e.g. PNG; though it does not address the multiple channels issue). On the other hand, it is likely that there are also users who have 3D input image data, but only single-slice segmentations. Those are also really 2D shapes (even though the real-world shape is 3D), but there it is more likely that dimensionality of the input data is 3D.

By the way, the latter also holds true for 2D DICOM (e.g. x-ray, mammography). ITK will read this in as 3D data, with size 1 in a certain dimension.

JoostJM:
> I would say the input is "truly 2D" if the ITK image passed as input is a 2D image, or perhaps if the spacing in z dimension is 0. Otherwise the definition can be ambiguous.

Very true, and would nicely help those users that use 2D input formats (e.g. PNG; though it does not address the multiple channels issue). On the other hand, it is likely that there are also users who have 3D input image data, but only single-slice segmentations. Those are also really 2D shapes (even though the real-world shape is 3D), but there it is more likely that dimensionality of the input data is 3D.

By the way, the latter also holds true for 2D DICOM (e.g. x-ray, mammography). ITK will read this in as 3D data, with size 1 in a certain dimension.

JoostJM:
> I would say the input is "truly 2D" if the ITK image passed as input is a 2D image, or perhaps if the spacing in z dimension is 0. Otherwise the definition can be ambiguous.

Very true, and would nicely help those users that use 2D input formats (e.g. PNG; though it does not address the multiple channels issue). On the other hand, it is likely that there are also users who have 3D input image data, but only single-slice segmentations. Those are also really 2D shapes (even though the real-world shape is 3D), but there it is more likely that dimensionality of the input data is 3D.

By the way, the latter also holds true for 2D DICOM (e.g. x-ray, mammography). ITK will read this in as 3D data, with size 1 in a certain dimension.

JoostJM:
> I would say the input is "truly 2D" if the ITK image passed as input is a 2D image, or perhaps if the spacing in z dimension is 0. Otherwise the definition can be ambiguous.

Very true, and would nicely help those users that use 2D input formats (e.g. PNG; though it does not address the multiple channels issue). On the other hand, it is likely that there are also users who have 3D input image data, but only single-slice segmentations. Those are also really 2D shapes (even though the real-world shape is 3D), but there it is more likely that dimensionality of the input data is 3D.

By the way, the latter also holds true for 2D DICOM (e.g. x-ray, mammography). ITK will read this in as 3D data, with size 1 in a certain dimension.

JoostJM:
> I would say the input is "truly 2D" if the ITK image passed as input is a 2D image, or perhaps if the spacing in z dimension is 0. Otherwise the definition can be ambiguous.

Very true, and would nicely help those users that use 2D input formats (e.g. PNG; though it does not address the multiple channels issue). On the other hand, it is likely that there are also users who have 3D input image data, but only single-slice segmentations. Those are also really 2D shapes (even though the real-world shape is 3D), but there it is more likely that dimensionality of the input data is 3D.

By the way, the latter also holds true for 2D DICOM (e.g. x-ray, mammography). ITK will read this in as 3D data, with size 1 in a certain dimension.

JoostJM:
> I would say the input is "truly 2D" if the ITK image passed as input is a 2D image, or perhaps if the spacing in z dimension is 0. Otherwise the definition can be ambiguous.

Very true, and would nicely help those users that use 2D input formats (e.g. PNG; though it does not address the multiple channels issue). On the other hand, it is likely that there are also users who have 3D input image data, but only single-slice segmentations. Those are also really 2D shapes (even though the real-world shape is 3D), but there it is more likely that dimensionality of the input data is 3D.

By the way, the latter also holds true for 2D DICOM (e.g. x-ray, mammography). ITK will read this in as 3D data, with size 1 in a certain dimension.

PR #445, opened 2018-12-07
Add support for overlapping masks / Slicer Segementation objects

JoostJM:
In the new version of Slicer, segmentation is done using Segmentation Objects, which allow for overlapping segmentation. When stored, these `.seg.nrrd` files can be read by SimpleITK as vector images. By extracting the nth index from the vector, a labelmap (3D volume) can be obtained representing the nth segement. To ensure correct functionality in the rest of PyRadiomics, all non-zero in the obtained mask are then set to the label value.

### DIFF
--- a/radiomics/imageoperations.py
+++ b/radiomics/imageoperations.py
@@ -11,6 +11,42 @@ from six.moves import range
 logger = logging.getLogger(__name__)
 
 
+def getMask(mask, **kwargs):
+  """
+  Function to get the correct mask. Includes enforcing a correct pixel data type.
+
+  Also supports extracting the mask for a segmentation (stored as SimpleITK Vector image) if necessary.
+  In this case, the mask at index ``label - 1`` is extracted. All non-zero pixels in that mask are then set to the
+  value of the label.
+
+  :param mask: SimpleITK Image object representing the mask. Can be a vector image to allow for overlapping masks.
+  :param kwargs: keyword arguments. If argument ``label`` is present, this is used to select the label. Otherwise
+    label ``1`` is assumed.
+  :return: 3D mask with pixel type UInt32
+  """
+  global logger
+  label = kwargs.get('label', 1)
+  if 'vector' in mask.GetPixelIDTypeAsString().lower():
+    logger.debug('Mask appears to be a segmentation object (=stored as vector image).')
+    n_components = mask.GetNumberOfComponentsPerPixel()
+    assert label <= n_components, \
+        "Mask %i requested, but segmentation object only contains %i objects" % (label, n_components)
+
+    logger.debug('Extracting mask at index %i', label - 1)
+    selector = sitk.VectorIndexSelectionCastImageFilter()
+    selector.SetIndex(label - 1)
+    mask = selector.Execute(mask)
+
+    logger.debug('Setting all non-zero elements in extracted mask to label value %i', label)
+    mask = sitk.Cast(mask != 0, sitk.sitkUInt32)  # Set all non-zero elements to 1, cast to UInt32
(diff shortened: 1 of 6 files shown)


radiomics/imageoperations.py
```diff
@@ -11,6 +11,42 @@
...
 
+def getMask(mask, **kwargs):
+  """
+  Function to get the correct mask. Includes enforcing a correct pixel data type.
+
+  Also supports extracting the mask for a segmentation (stored as SimpleITK Vector image) if necessary.
+  In this case, the mask at index ``label - 1`` is extracted. All non-zero pixels in that mask are then set to the
+  value of the label.
+
+  :param mask: SimpleITK Image object representing the mask. Can be a vector image to allow for overlapping masks.
```
fedorov: Do I understand it correctly that `label` argument has dual meaning - to set the value of the label that should be used in the mask, when single volume, and to set the vector component otherwise? If this is the case, I think this is confusing. I would introduce additional argument to select the vector component.

JoostJM: Correct, although I considered it more from the viewpoint of Slicer's Segmentation/Labelmap objects, where ROIs are defined either by their index in the vector image (Segmentation, supports overlap) __or__ the pixel value (Labelmap, does not support overlap).

In this setting, the `label` parameter just defines which ROI you want, regardless of the method you used to store it (Segmentation or Labelmap).

fedorov: Yes, I see. The issue is, there is probably no way to know how the input vector image was created, and whether this assumption holds. It's your call!

fedorov: I think in practice, either way is probably fine, since I don't think this format is used widely. I just think that it is confusing to overload parameters and give them implicit meaning in specific context. We also don't know whether conventions used currently in Slicer Segmentations will not change over time, or whether Slicer developers can write code that would create programmatically `.seg.nrrd` that does not follow this convention. Overall, I think it is cleaner to separate parameters for channel selection and for the label within channel. We could also have a default setting that channel is set to be identical to label if not specified, but leave the flexibility to adjust the channel if needed. But this is just my thoughts. I am fine if you proceed with your current implementation.

pieper: It's a good point that the '.seg.nrrd' convention is ad hoc and could change without notice.  We should at least suggest that there be a version number added.  I'm not even sure it's documented anywhere.

Having an independent flag for channel selection and label within channel also makes sense.  Then a shorthand 'seg.nrrd' mode would set the values according to segment editor conventions.

Throwing exceptions or at least warnings for any unexpected input would also be good (e.g. a vector image that isn't a seg.nrrd isn't likely to make sense as a label definition).

JoostJM:
@fedorov, @pieper,

I was reviewing this PR, and I think (at least for now) I prefer to remove any ambiguities. To that end, I'll update this PR, so that vector images as masks are supported, and the correct vector index is selected by `label_idx`, which will be available from the commandline (as it's just another setting, like `label`). The only real addition is that, besides the optional column `Label`, optional column `Label_idx` will also be recognized in the input csv (batch mode), and correspond to the `label_idx` setting, similar to the link between `Label` column and `label` setting.

Inside the feature classes, the mask is still required to be a 3D SimpleITK image with scalar values (i.e. the correct volume in the mask vector is selected prior to being passed to the feature class, as it is in the current version of this PR).

fedorov:
@JoostJM sounds good! The only comment is that I personally don't like the name `label_idx` - I would suggest something more descriptive, maybe `label_channel_index`?

JoostJM:
@fedorov Sounds good, although `label_channel_index` is quite long. How about `label_channel`?

PR #481, opened 2019-04-17
ENH: Refactor command line and feature extractor

JoostJM:
Commandline
- Instantiate feature extractor in main loop and pass it to the worker processes (this will only load the parameter file once, including all checks etc.)
- Simplify construction of `caseGenerator` (only contains case index and case, setting(overrides) are no longer needed as these are used to instantiate the extractor)
- Remove function '_buildGenerator', as construction of the generator is now a one-line code. Deprecated commandline switch `label` is now handled inside the `_parseOverrides` function, which in turn is called in `_processCases`.

FeatureExtractor
- RENAME CLASS FROM `RadiomicsFeaturesExtractor` TO `RadiomicsFeatureExtractor`, which shows better correspondence to the module name.
- Ensure that the feature extractor instance can be sent over a Pipe using pickle:
  - logger object is now defined as a global variable, not a variable of the extractor instance. This is done because this object often contains stream objects, which cannot be pickled
  - define geometryTolerance as a global variable, as this will then be set separately for each worker thread, as necessary. This prevents the tolerance from being set once in the parent process, but not in the child-processes.
  - Reduce the number of instance variables in the extractor as much as possible. Current variables retained (and pickled) are `settings`, `enabledImagetypes` `enabledFeatures`, `featureClassNames`
- Move parsing of shape-calculation settings to the function `computeShape()`
- Update docstring of `enableImageTypesByName` to reflect additions to PyRadiomics' filters
- Move the binCount warning to the instantiation of the extractor, resulting in 1 warning overall, instead of 1 warning for each extraction case.

Apply changes to rest of PyRadiomics' code to reflect changes made in feature extractor API.


fedorov:
@JoostJM I believe this branch is now in conflict, because I made few small fixes in eccdbc28d62d52ab59f1830b265e439ecbb17c0c (invalid variable name was used to write JSON output) and a542f691ccf5064717bd6e2d0b5762fe453c68dc (added indentation to the JSON output) to the pyradiomics script. 

I rebased and resolved locally, but I feel very hesitant to force-push to your branch, so I will leave it to you to resolve.

fedorov:
Also, is this branch ready for testing? I actually need the batch extraction capability for a project (which we should discuss next Wednesday, by the way!) - that's how I came across the issues I fixed in the aforementioned commits.

JoostJM:
@fedorov I'll rebase and merge.

Be aware that this PR also contains a name change to the feature extractor class (to make it more consistent with the module name)

PR #492, opened 2019-06-11
BUG: Mitigate integer overflow

maekclena:
In `cshape.c`, `v_max` would quickly overflow with big images resulting in messages like `RuntimeError: Calculation of Shape coefficients failed`. To mitigate this effect, this PR changes its type from `int` to `size_t` (usually `unsigned long`).


JoostJM:
Thanks for the contribution :+1:

PR #527, opened 2019-09-04
Add Docker for PyRadiomics-dcm Labs

JoostJM:
Add docker, based on PyRadiomics:CLI docker and latest QIICR/dcmqi docker image to containerize DICOM -> DICOM workflow


JoostJM:
Additionally updates the CLI dockerfile in 2 respects:

- Build PyRadiomics from the build context instead of cloning it from git. This ensures the correct version of PyRadiomics is used when building the docker.
- Include bzip2-devel before compiling Python. This is required to build module `bz2`, which in turn is required by pandas, which is used in the labs docker.

Issue #544, opened 2019-12-18
[BUG] When resampling pixel spacing


KathrynSch:
**Describe the bug**
Using the feature extractor on a 2D image with 2D mask raises a `ValueError` in the `resampleImage` of  `imageoperations.py`: 
`could not broadcast input array from shape (3) into shape (2)`.
This error is raised when the image and mask original pixel spacing equals to the `resampledPixelSpacing` defined in the feature extractor settings. 
The image and mask are of size (512, 512), mask contains integers [0, 1].

**PyRadiomics configuration**
The feature extractor has been initialized with the following settings:
```
settings = {
    'minimumROIDimensions': 2,
    'minimumROISize': None,
    'normalize': False,
    'normalizeScale': 1,
    'removeOutliers': None,
    'resampledPixelSpacing': [0.75, 0.75],
    'interpolator': 'sitkBSpline',
    'preCrop': False,
    'padDistance': 5,
    'distances': [1],
    'force2D': False,
    'force2Ddimension': 0,
    'resegmentRange': None,
    'label': 1,
    'additionalInfo': True}
``` 

**To Reproduce**
```
import SimpleITK as sitk
import radiomics
import numpy as np

# initialize feature extractor with default settings except resampledPixelSpacing
settings = {
    'minimumROIDimensions': 2,
    'minimumROISize': None,
    'normalize': False,
    'normalizeScale': 1,
    'removeOutliers': None,
    'resampledPixelSpacing': [0.7, 0.7],
    'interpolator': 'sitkBSpline',
    'preCrop': False,
    'padDistance': 5,
    'distances': [1],
    'force2D': False,
    'force2Ddimension': 0,
    'resegmentRange': None,
    'label': 1,
    'additionalInfo': True}
extractor = radiomics.featureextractor.RadiomicsFeatureExtractor(**settings)

# params to raise error
shape = (512, 512)
original_pixel_spacing = [0.7, 0.7]

# generate dummy data
image = np.random.rand(shape[0], shape[1])
mask = np.zeros(shape) # init mask with 0
mask[:10, :10] = np.ones((10, 10)) # simulate ROI

# convert image and mask to sitk images and set pixel spacing
image_sitk = sitk.GetImageFromArray(image)
image_sitk.SetSpacing(original_pixel_spacing)
mask_sitk = sitk.GetImageFromArray(mask)
mask_sitk.SetSpacing(original_pixel_spacing)

# extract radiomics
[...]

PR #545, opened 2019-12-30
Fix resampling operation for 2d images

KathrynSch:
#544
cc: @Radiomics/developers
Note: Tests fail but I ran them on the current master branch and they fail as well


KathrynSch:
cc: @Radiomics/developers

KathrynSch:
Tests fail but I ran them on the current master branch and they fail as well

JoostJM:
@KathrynSch, Thanks!

The test do indeed fail, I'm currently looking into why. It's not due to your change, so I'm merging anyway.

Issue #456, opened 2019-01-30
Voxel-wise feature calculation?


Kuttner:
Hi,
I am trying to calculate a parametric image by extracting voxel-wise features from a square kernel. After reading the reply in this thread: https://groups.google.com/forum/#!topic/pyradiomics/k6WXuwYTqNA
I tried to run the helloVoxel.py from: https://github.com/Radiomics/pyradiomics/blob/master/examples/helloVoxel.py
using the following parameter file (exampleVoxel.yaml):
______________________________________
imageType:
  Original: {}

featureClass:
  firstorder:
    - 'Mean'

setting:
  binWidth: 25
  force2D: true
  label: 1

voxelSetting:
  kernelRadius: 2
  maskedKernel: true
  initValue: nan
______________________________________

I.e. I would like to calculate the mean inside a 5x5 kernel for every pixel inside the mask. 

However, the output is just constant value of 7.60209 inside the mask, throughout the entire volume (and of course nan outside the mask). And after trying out with simple synthetic images, it seems that it returns the mean inside the entire mask for every voxel in the mask, and not the mean inside the 5x5 kernel for every voxel inside the mask.

Could you please look into if this is a bug, or if I made a mistake here?

Thank you for your kind help!

Regards
Samuel Kuttner

JoostJM:
@Kuttner That is indeed a bug, good catch. I fixed it in 81e713a (the current master).
If you can build the latest PyRadiomics master from source, the issue should be fixed. If you want to use the pre-built wheels, the fix will be included in the next release.

Kuttner:
@JoostJM Thank you for looking into this. I think I managed to upgrade. Running: "pyradiomics --version" gives the output: "pyradiomics 2.1.2.post21+gb92bd22". Is this correct?

However, still, running the helloVoxel.py, with the above *.yaml file just gives constant value of 7.60209 inside the mask. Could you provide me with a minimal working example that shows it is working for you?

Thanks a lot!

JoostJM:
Thats the latest master, the same I ran my test with. As a small bonus, this also contains a CLI interface for voxelwise extraction.

From the examples folder, can you run the following? 

- open and edit `exampleVoxel.yaml` so it looks like this:
```

imageType:
  Original: {}

featureClass:
  firstorder:
    - Mean

setting:
  binWidth: 25
  force2D: true
  label: 1

voxelSetting:
  kernelRadius: 2
  maskedKernel: true
  initValue: nan

```
[...]

PR #602, opened 2020-07-02
BUG: Fail gracefully when grayvalue <= 0

JoostJM:
"0" is an invalid value for gray level and causes an index error in the calculation of texture matrices (e.g. for GLSZM, correct index in matrix is calculated as (i, j) = (grayvalue - 1, region size), which in case of gray value 0 results in index -1, which is invalid).

In some case, this resulted in a graceful error where an IndexError is raised, but this is most likely because the datatype used for the index (`size_t`) does not allow for negative indices on some systems and therefore results in an overflow, which on the subsequent check is then indeed larger than the maximum allowed index (which is checked). In other cases, `size_t` datatype does allow for negative values and therefore passes the maximum index check and subsequently fails with a Process terminated.

Add an additional check to ensure the gray level is > 0, otherwise, fail the C Extension gracefully. Non-gray level properties (i.e. run length, zone size, dependence counts) are calculated in the C extension and ensured to be >= 0.

### DIFF
--- a/radiomics/src/cmatrices.c
+++ b/radiomics/src/cmatrices.c
@@ -77,7 +77,7 @@ int calculate_glcm(int *image, char *mask, int *size, int *bb, int *strides, int
         if (j != i && mask[j])
         {
           glcm_idx = a + (image[j]-1) * Na + (image[i]-1) * Na * Ng;
-          if (glcm_idx >= glcm_idx_max)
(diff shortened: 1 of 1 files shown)


Kuttner:
@JoostJM, thanks for this fix. Relating back to my previous issue #456, I would need to be able to also implement dynamic binning. Is it possible to merge these two fixes (#456 and #602) into the next release? Thanks!

PR #636, opened 2020-09-05
ENH: Add optional progress reporting for voxel-based extraction

JoostJM:
Update base.py to add progress reporting when computing voxel batches. Uses `radiomics.getProgressReporter`, which should be set to something like `tqdm.tqdm` to work. Only shows progress bar when logging level is at least INFO (so not when WARNING or higher).


JoostJM:
Merging as the error in Travis CI is not due to changes in this PR.

PR #644, opened 2020-09-30
DOCS: Add force2D setting to example settings

JoostJM:
In the example showing settings for 2D MR extraction, the setting `force2D` was missing. Add this setting with a value of `true` to ensure this example forces a 2D extraction.

@rcuocolo


rcuocolo (approved):
Added missing force2D option.

rcuocolo (approved):

Issue #653, opened 2020-10-20
[BUG] numpy required in setup.py


davide84:
**Describe the bug**
Module numpy is imported in setup.py. This causes "pip install pyradiomics" to fail if numpy is not yet installed. The issue implies that pyradiomics can not be listed as a project dependency in requirements.txt together with numpy, but numpy must be previously installed. This is e.g. the case of fresh installations.

**PyRadiomics configuration**
Default - no customization.

**PyRadiomics log file**
None

**To Reproduce**
"pip install pyradiomics" from a fresh environment with no numpy installed.

**Expected behavior**
Numpy should be pulled as dependency when installing pyradiomics and not halting the installation.

**Version (please complete the following information):**
Python 3.8 under Ubuntu 20.04, pyradiomics 3.0.1

**Additional context**
None

davide84:
It seems to be a common problem of packages depending on numpy.

There might be a proper fix to this type of issues, apparently addressed with [PEP 518;](https://www.python.org/dev/peps/pep-0518/) an example is here: https://github.com/SciTools/cartopy/issues/1112#issuecomment-418400367

JoostJM:
This is indeed the case, though it only applies to the case where PyRadiomics needs to be built from source (for pre-distributed wheels, setup.py is not executed on install, but rather when the wheel is built).

PyRadiomics needs numpy to be installed, as the C-headers are needed to compile the C-extensions.

The suggestion in your second comment may help, but as mentioned in that thread, can cause some unexpected behaviour (like pip re-installing numpy if it's already installed). This has the potential to break stuff in special environments (like 3D slicer), so I don't wan't to add it to PyRadiomics just yet.

We did, however, add a 'setup-requires.txt' file which is also passed in the setup function. The problem is that `setup.py` tries to import numpy, and therefore does not have the chance to indicate numpy should be installed first.

I'll run some tests to see if I can fix this problem in some other way.

psteinb:
This is falling on my feet in a project that I am working on, i.e. my CI pipelines crashes when including pyradiomics. 
In other words: I'd like to install `pyradiomics` through pip+`requirements.txt` with python 3.9, but the numpy problem holds me back.

[...]

PR #761, opened 2022-04-08
Defer numpy import on first pass of setup.py

masalim2:
This PR addresses #653  by ensuring that `numpy` is not required on the first pass of `setup.py`.  This makes `pyradiomics` easy to install into new virtual environments without having an a priori installation of numpy.

The update is based on the fact that setuptools invokes `setup.py` multiple times, and on the first pass, you must guard import statements of modules in the `setup_requires` section.   We can therefore mock out the `numpy` module on the first pass, which should not affect the subsequent installation steps after the build system has fetched numpy and installed it.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,13 @@
 from distutils import sysconfig
 import platform
 
-import numpy
+try:
+    import numpy
+except:  # noqa: E722
+    from unittest.mock import MagicMock
+    sys.modules["numpy"] = MagicMock()
+    import numpy
+
 from setuptools import Extension, setup
 import versioneer
 


masalim2:
Current CI failures are due to flake8 linting of labs/pyradiomics-dcm/pyradiomics-dcm.py -- unrelated to this change

JoostJM:
Can you rebase on PR #788? This allows me to see if your changes do not break the build process.

masalim2:
@JoostJM All set -- this PR is now just one commit ahead of `JoostJM/circle-ci-mac-os`

JoostJM:
I have only recently finished the updates to get CI working again. Involved not only a transition to different CI servers, but also a transition of testing package...

I'm about to make a new release (3.1.0). If you can rebase one more time on the new release (should not break much), then I'll create release 3.1.1 once this PR is merged. I don´t want to wait with release 3.1.0, as it is already long overdue.

JoostJM:
In addition, as you can see in #807, adding the pyproject.toml also fixes this particular error. I think your issue will therefore be resolved upon the 3.1.0 release. In any case, I'm still planning to include this PR in a 3.1.1 release.

hjmjohnson:
Issue resolved by migrating to pyproject.toml infrastructure.  Removing the setup.py file.

PR #826, opened 2023-05-18
CI: Add MacOS Conda deployment step

JoostJM:
Recently, CI for MacOs was moved to CircleCI. Wheels intended for PyPi distribution are build as part of the general testing, including wheels for MacOS. However, conda deployment is handled in a separate step, which in CircleCI currently only handles distribution of Linux Packages. Therefore, add a conda-deployment step to generate and upload pyradiomics packages for conda on MacOS.


hjmjohnson:
@JoostJM I don't this is is relevant any longer.  I am closing based on the assumption that only github-actions are relevant.

Issue #831, opened 2023-06-05
[BUG]pyradiomics=3.1.0 failed on linux with python=3.7.16


lyhyl:
**Describe the bug**
The lastest pyradiomics=3.1.0 failed on linux, but good on windows.

**PyRadiomics configuration**
N/A

**PyRadiomics log file**
```
Error loading C extensions
Traceback (most recent call last):
  File "/home/shery/downloads/anaconda3/yes/envs/pyrad/lib/python3.7/site-packages/radiomics/__init__.py", line 286, in <module>
    from radiomics import _cmatrices as cMatrices  # noqa: F401
ImportError: libpython3.7m.so.1.0: cannot open shared object file: No such file or directory
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/shery/downloads/anaconda3/yes/envs/pyrad/lib/python3.7/site-packages/radiomics/__init__.py", line 297, in <module>
    raise e
  File "/home/shery/downloads/anaconda3/yes/envs/pyrad/lib/python3.7/site-packages/radiomics/__init__.py", line 286, in <module>
    from radiomics import _cmatrices as cMatrices  # noqa: F401
ImportError: libpython3.7m.so.1.0: cannot open shared object file: No such file or directory
```

**To Reproduce**
1. conda create -n pyrad python=3.7
2. conda activate pyrad
3. pip install pyradiomics
4. python
5. import radiomics

**Expected behavior**
import radiomics correctly without error

**Version (please complete the following information):**
 - OS: ubuntu
 - Python version: 3.7.16
 - PyRadiomics version 3.1.0

**Additional context**
maybe related to https://github.com/AIM-Harvard/pyradiomics/issues/828 https://github.com/AIM-Harvard/pyradiomics/issues/829?

lyhyl:
Here need to be update.
https://github.com/AIM-Harvard/pyradiomics/blob/6a761c4eb01905cfcb1139b0b158547c55edd4df/pyproject.toml#L7

rcuocolo:
I'm also having issues updating from 3.0.1 to 3.1.0 on Ubuntu 22.04, similar build errors as those mentioned here. Python 3.10.2

agr78:
With Python 3.7.16 on Ubuntu 18.04, the following worked for me:

`sudo apt-get install libpython3.7`

[...]

Issue #828, opened 2023-05-22
[BUG] pypi version mismatch


jxchen01:
**Describe the bug**
when installing the package via pypi, there is a meta data mismatch:
```
Requested pyradiomics from https://files.pythonhosted.org/packages/03/c1/20fc2c50ab1e3304da36d866042a1905a2b05a1431ece35448ab6b4578f2/pyradiomics-3.1.0.tar.gz has inconsistent version: expected '3.1.0', but metadata has '3.0.1a1'
```

**PyRadiomics configuration**

N/A

**To Reproduce**

`pip install pyradiomics` with Python=3.9

**Version (please complete the following information):**
 - OS: MacOS 13.3.1
 - Python version: 3.9
 - PyRadiomics version: 3.1.0

jxchen01:
further error message:
```
 Using cached pyradiomics-3.0.1.tar.gz (34.5 MB)
  Preparing metadata (setup.py) ... done
Requirement already satisfied: numpy>=1.9.2 in ./opt/anaconda3/envs/validomics_dev/lib/python3.8/site-packages (from pyradiomics) (1.24.3)
Collecting SimpleITK>=0.9.1
  Downloading SimpleITK-2.2.1-cp38-cp38-macosx_10_9_x86_64.whl (45.5 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 45.5/45.5 MB 9.1 MB/s eta 0:00:00
Collecting PyWavelets>=0.4.0
  Using cached PyWavelets-1.4.1-cp38-cp38-macosx_10_13_x86_64.whl (4.3 MB)
Collecting pykwalify>=1.6.0
  Using cached pykwalify-1.8.0-py2.py3-none-any.whl (24 kB)
Collecting six>=1.10.0
  Using cached six-1.16.0-py2.py3-none-any.whl (11 kB)
Collecting ruamel.yaml>=0.16.0
  Using cached ruamel.yaml-0.17.26-py3-none-any.whl (109 kB)
Collecting docopt>=0.6.2
  Using cached docopt-0.6.2.tar.gz (25 kB)
  Preparing metadata (setup.py) ... done
Collecting python-dateutil>=2.8.0
  Using cached python_dateutil-2.8.2-py2.py3-none-any.whl (247 kB)
Collecting ruamel.yaml.clib>=0.2.7
[...]

PR #836, opened 2023-08-14
Bump version in pyproject toml to avoid metadata clash with file version

sjswerdloff:
A few of the issues ( #831 and  #828 ) posted appear to have root cause in the metadata clash with the package publish/file version.
This is an attempt at resolving that by updating the version value in the pyproject.toml file.
There may be other causes: poetry complains about pep-517 compliance.
reverted some of my own commits so that this only alters pyproject.toml and leaves the versioner changes to the CI/CD (if that's how things work).  there were build failures because of changes made by versioneer (and I don't have pre-commit hooks set up that match the CI expectations).

The published version claims 3.1.0, so the next valid semantic version is 3.1.1


sjswerdloff:
@JoostJM do you think I'm on the right track with this, or is there a lot more involved in the versioning and meta-data?

pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "pyradiomics"
-version = "3.0.1a1"
+version = "3.1.1"
 authors = [
     { name = "PyRadimiomics Community", email = "pyradiomics@googlegroups.com"}
```
fedorov, line 9:
```suggestion
    { name = "PyRadiomics Community", email = "pyradiomics@googlegroups.com"}
```

PR #847, opened 2023-11-07
added Dicom file reader

ManuelaRosina:
Extension of the loadImage function. If the given filepath is a folder the simpleITK ImageSeriesReader is used to load the DICOM files in the folder.


hjmjohnson:
Updated in issue #919.

PR #857, opened 2023-12-06
SafeLoadRemove Error in Slicer

DanXieJY:
I recently found the SlicerRadiomics extension does not work in 3D Slicer and below is the issue that I decribed in.
https://github.com/AIM-Harvard/SlicerRadiomics/issues/80

I tried to modify the code with the help of @pieper :
1. Add the code below to the top of the file:
`from ruamel.yaml import YAML`
2. Add the code below before line 353:
`yaml = YAML(typ='safe', pure=True) `
3. Change what was in line 353 to:
`settingsSchema = yaml.load(schema)['mapping']['setting']['mapping']`

### DIFF
--- a/radiomics/scripts/__init__.py
+++ b/radiomics/scripts/__init__.py
@@ -19,6 +19,7 @@ import radiomics
 import radiomics.featureextractor
 from . import segment, voxel
 
+from ruamel.yaml import YAML
 
 class PyRadiomicsCommandLine:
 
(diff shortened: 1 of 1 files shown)


pieper:
I've tested this manually on linux and it does fix the runtime issue.  @JoostJM or @fedorov can you merge it?  I'm not sure what's wrong with circleci but it looks like the ci code has the same issue that this PR fixes in pyradiomics itself and I don't think it should block this PR.

PR #884, opened 2024-06-25
fix typo in gldm sde formula

standakozak:
Fixed the formula for SDE calculation from GLDM. The present formula is actually for LGLE.
I also checked the formula against [IBSI documentation](https://arxiv.org/pdf/1612.07003) (Chapter 3.11.1).

### DIFF
--- a/radiomics/gldm.py
+++ b/radiomics/gldm.py
@@ -133,7 +133,7 @@ class RadiomicsGLDM(base.RadiomicsFeaturesBase):
     **1. Small Dependence Emphasis (SDE)**
 
     .. math::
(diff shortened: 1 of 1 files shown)


L1cardo:
Will this affect much on data analysis? I have done a lot of work already

standakozak:
Hi @L1cardo,
no, this does not affect the implementation. It just fixes a typo in the documentation.

L1cardo:
> Hi @L1cardo, no, this does not affect the implementation. It just fixes a typo in the documentation.

Thanks for the reply! ❤️

PR #908, opened 2025-05-13
Full update to require python 3.9 and other goodies

hjmjohnson:
Update code style and enforce style consistency with pre-commit hooks.


hjmjohnson:
Fixed in other PRs

Issue #883, opened 2024-06-25
[FEAT EXTRACTION] Slow voxel-based features extraction and 2 solutions (CPU & GPU/PyTorch).


lyhyl:
**Description**
I use PyRadiomics to extract voxel-based features and noticed that it is extremely slow, especially GLCM features, like https://github.com/AIM-Harvard/pyradiomics/issues/679.
I use pyinstrument to profile extraction and I found that `numpy.delete` and `ndarray.copy` takes lots of time.
![image](https://github.com/AIM-Harvard/pyradiomics/assets/3370369/8178b1db-e02b-4101-bc00-23e0b15246d4)

so I try to speed up and made a PR(https://github.com/AIM-Harvard/pyradiomics/pull/882).
After this modification, I got improvement of 14s:
![image](https://github.com/AIM-Harvard/pyradiomics/assets/3370369/7016b06f-5811-4193-81d7-e71a7c991d2d)

with very small error:
```
all close: True
max abs diff: 2.7939677238464355e-09
max rel diff1: 3.6155523019942848e-12
max rel diff2: 3.6156633242967473e-12
```

**PyRadiomics configuration**
NA

**PyRadiomics log file**
```
batch: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 14/14.0 [10:31<00:00, 45.13s/it]
batch: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 14/14.0 [10:18<00:00, 44.16s/it]
```

**To Reproduce**
```python
import logging
import warnings
from typing import List, Type

import numpy as np
import radiomics
import radiomics.base
import SimpleITK as sitk
import torch
from pyinstrument import Profiler
from radiomics import glcm
from radiomics.featureextractor import RadiomicsFeatureExtractor
from radiomics.imageoperations import checkMask, cropToTumorMask
from tqdm import tqdm

from RadiomicsGLCM2 import RadiomicsGLCM2


def get_default_features():
    return ['Autocorrelation', 'JointAverage', 'ClusterProminence', 'ClusterShade',
            'ClusterTendency', 'Contrast', 'Correlation', 'DifferenceAverage', 'DifferenceEntropy',
            'DifferenceVariance', 'JointEnergy', 'JointEntropy', 'Imc1', 'Imc2',
[...]

PR #882, opened 2024-06-25
improve glcm performance (voxel-based features extraction)

lyhyl:
https://github.com/AIM-Harvard/pyradiomics/issues/883

PR #918, opened 2025-05-18
PERF: Improve glcm performance

hjmjohnson:
Addresses issues identified in #883 to reduce the number of numpy copy and delete operations.  Re-implements issue #882.


lyhyl:
Thank you for your contributions to the pyradiomics module, making it more compatible with modern Python module management (new project structure works so smoothly when set up with uv). I've tested the new implementation (Linux, Python 3.12) and confirmed that the relevant tests have passed. Additionally, I think it would be helpful to add a brief comment explaining the reason for this change, as it might make things clearer (PR #920). Is there anything else I can assist with?
image

PR #919, opened 2025-05-18
ENH: If file path is directory, attempt to read dicom series

hjmjohnson:
Replaces issue #847: added Dicom file reader

code updated to match current coding style formats.


hjmjohnson:
This is an extremely low negative impact change.

Issue #954, opened 2026-07-12
Cut a release from master: PyPI still serves the May 2023 wheel set


vitalii2077:
## What users hit today

`pip install pyradiomics` on a current Python does not produce a working install.

The newest release on PyPI is 3.1.0, published 17 May 2023. That is its complete wheel set:

```
pyradiomics-3.1.0-cp37-cp37m-{macosx_11_0_x86_64, manylinux1_x86_64, win_amd64}.whl
pyradiomics-3.1.0-cp38-cp38-{macosx_11_0_x86_64, manylinux1_x86_64, win_amd64}.whl
pyradiomics-3.1.0-cp39-cp39-{macosx_11_0_x86_64, manylinux1_x86_64, win_amd64}.whl
pyradiomics-3.1.0.tar.gz
```

Nothing for CPython 3.10 or newer, and no macOS arm64 wheel at all. Two details make this worse than a plain "no wheel yet":

- The 3.1.0 release carries an **empty `Requires-Python`** on PyPI. pip therefore does not skip it on 3.12 or 3.13. It selects the release, finds no matching wheel, falls back to the sdist, and tries to build from source, which fails. That is the failure reported in #949, open since April with eight comments and no maintainer reply.
- The 3.1.0 sdist reports its version as `3.0.1a1` in package metadata, which breaks strict resolvers outright. That is #933, open since July 2025 with no replies at all. The reporter is explicitly asking for a release: "This has been fixed in the master branch, but hasn't been pushed into a release. Can that be done so installing from pypi works again?"

The practical result is that downstream projects pin an old Python, pin `numpy<2`, or vendor a source-build toolchain, purely to keep pyradiomics installable.

## master has been releasable for well over a year

The build modernization is already merged and sitting on `master`:

- scikit-build-core replaced setuptools in cb87a3e, 28 January 2025.
- The cibuildwheel build selector was tuned in 1e6644b, 17 May 2025.
- `pyproject.toml` declares `requires-python = ">=3.9"`, and `.github/workflows/cd.yml:55` builds through cibuildwheel v3.0.

None of that has ever reached a user, because no release has been cut since the migration.

## Why no release has shipped: the publish job targets TestPyPI

`.github/workflows/cd.yml:94-98`:

```yaml
      - uses: pypa/gh-action-pypi-publish@release/v1
        with:
          # Remember to tell (test-)pypi about this repo before publishing
          # Remove this line to publish to PyPI
[...]

PR #955, opened 2026-07-12
COMP: Publish releases to PyPI instead of TestPyPI

vbabenko97:
## Summary

The publish step in `.github/workflows/cd.yml` pinned `repository-url` to `https://test.pypi.org/legacy/`, so every release published from this repository uploads to TestPyPI. Production PyPI therefore still serves only the cp37 to cp39 wheels from 3.1.0 (May 2023), which do not install on a current Python. This removes that pin, so `pypa/gh-action-pypi-publish` falls back to its default target, PyPI.

The line sat directly under a comment reading `# Remove this line to publish to PyPI`, so this is the change that comment was asking for.

The wider release picture is in #954: why nothing modernized has reached PyPI since May 2023, and the trusted-publisher step that only a maintainer can complete. This pull request is the small, self-contained piece of that.

## Why the diff removes four lines rather than one

`repository-url` was the only key under `with:`. Removing just that key leaves an empty `with:` mapping, which the GitHub Actions workflow schema rejects. This repository's own `check-github-workflows` pre-commit hook catches it:

```
$.jobs.upload_all.steps[2].with: None is not of type 'object'
```

So the `with:` block goes with the key. This is still a single functional change: the publish step is now an unconfigured `uses:` and nothing else in the workflow moves.

## The prerequisite this PR cannot satisfy

The deleted block also carried the comment `# Remember to tell (test-)pypi about this repo before publishing`. That prerequisite is real and still outstanding, so it is worth restating here rather than losing it with the comment.

Publishing needs a **trusted publisher (OIDC) entry** on the `pyradiomics` project on PyPI, pointing at owner `AIM-Harvard`, repository `pyradiomics`, workflow `cd.yml`, environment `pypi`. Only a PyPI project owner can add that. The workflow half is already in place: `upload_all` requests `id-token: write` and runs in the `pypi` environment (`.github/workflows/cd.yml:74-78`).

Without that entry, merging this PR is harmless but inert. The first release after it will fail at the upload step rather than publish to the wrong index.

## A note on the CI status of this PR

`cd.yml` runs on any pull request that touches `cd.yml` (`.github/workflows/cd.yml:8-10`), so this PR triggers the `wheels` workflow. It may come back red, and that would not be caused by this change.

The workflow has not had a green run since 16 June 2025. Its last five runs, between 28 July and 27 October 2025, all failed `build_wheels`. The logs have since expired, so I cannot say what broke.

Those red runs do not show that master is broken, and it is worth being precise about why. `cd.yml` has no `push` trigger. It runs only on `workflow_dispatch`, on a published release, and on pull requests that touch `cd.yml` itself (lines 3-10). Every run in the workflow's history is therefore a Dependabot pull request, and each of those bumps the very actions the workflow depends on. The most recent, #942, moves `pypa/cibuildwheel` from v3.0 to v3.2, `actions/checkout` from v4 to v5, `actions/upload-artifact` from v4 to v5, and `actions/setup-python` from v5 to v6. Master pins the older versions, and that combination has never actually been run.

The honest position is that the state of the wheel build on master is unknown rather than known-broken, and it has not produced a green run in thirteen months.

Nothing can be published from this pull request in any case. `upload_all` is gated on `github.event_name == 'release' && github.event.action == 'published'` (line 80), so it stays skipped on every pull request.

## Observation, not part of this PR

While checking whether a release from master would actually produce modern wheels, I looked at the cibuildwheel selector at `pyproject.toml:104`:

```toml
build = "cp3*-manylinux_x86_64 cp3*-manylinux_aarch64 cp3*-win_amd64 cp3*-macosx* "
```

The `cp3*` wildcard names no explicit interpreter range, and the TestPyPI artifacts for `3.1.1a3` show only cp37 to cp39, which together look like evidence that the selector is stuck on old interpreters. I do not think it is, and I am flagging it as an observation rather than bundling a second change into this PR.

Those TestPyPI artifacts predate the build migration. Tag `v3.1.1a3` points at commit 67d1f26 from 20 May 2023, whose `pyproject.toml` still used `setuptools` and `versioneer`; the `manylinux1` platform tag on those files confirms they came from the old build. As positive evidence that the current configuration is fine, the `steubk/pyradiomics` fork builds cp310 through cp313 wheels for manylinux x86_64, macOS arm64 and win_amd64 from essentially this configuration.

If you want certainty before tagging, `cd.yml` already declares `workflow_dispatch` (line 4). Running the workflow manually builds the full wheel matrix and publishes nothing, which shows the real wheel set in one click.

## Verification

Validated the edited workflow against the same schema the repository's `check-github-workflows` hook uses:

```
$ check-jsonschema --builtin-schema vendor.github-workflows .github/workflows/cd.yml
ok -- validation done
```


vbabenko97:
This implements the workflow fix described in #954, which I filed from my other account.

